PR #1, opened 2021-04-12
feat: set up project skeleton for new CLI

### LINKED ISSUE #2: Bootstrap the project
Make it possible to develop in this repo using our usual flow.
* Set up go modules
* Pick a CLI library to use for now, and make the program entry-point with it
* Add enough build/test harness for CI to run

danxmoran:
Closes #2

* Set up go modules, using Go 1.16
* Add the entry-point for the CLI, with the `version` command implemented
* Add our standard linting tools/scripts
* Add a `Makefile` to be the source-of-truth for build/test/lint commands
* Add a basic CircleCI workflow


.circleci/config.yml
```diff
@@ -0,0 +1,48 @@
+version: "2.1"
+
+workflows:
+  version: 2
+  build:
+    jobs:
+      - lint
+      - build
+      - test
```
danxmoran, line 9: Eventually we might want a pre-step to cache dependencies. If we hit that point, we'll need to deal with golang/go#43994

cmd/influx/main.go
```diff
@@ -0,0 +1,44 @@
+package main
+
+import (
+	"fmt"
+	"os"
+	"time"
+
+	"github.com/urfave/cli/v2"
```
danxmoran, line 8: Risking an innovation token on `urfave/cli` over `spf13/cobra`. `cobra` has always felt more heavyweight/verbose to me than a CLI parser needs to be. `cli` looks very similar to the `kit/cli` convenience wrapper in `influxdb`.

If we hit a wall with it, we can fall back to using `cobra` and copy the code from `influxdb`.

lesam (approved):

PR #31, opened 2021-04-13
feat: implement ping command

danxmoran:
`ping` itself is very simple, but the implementation needed this pre-work:
* Create a (temporary) `api.yml` copy to model the subset of APIs covered by the CLI
* Set up `oapi-codegen` to generate type & client definitions from the YML
* Add support to the CLI for loading local config from disk
* Add global CLI flags
* Add the `ping` implementation
* Add tests for `ping`


.gitattributes
```diff
@@ -0,0 +1,2 @@
+api/types.gen.go  linguist-generated=true
+api/client.gen.go linguist-generated=true
```
danxmoran, line 2: Hides the generated files by default in GitHub diffs

api/api.yml
```diff
@@ -0,0 +1,113 @@
+openapi: 3.0.0
```
danxmoran, line 1: Eventually, we'll ref YML from our common API repo instead of maintaining a separate copy here. The copy is useful in the short-term to decouple the two projects.

api/types.go
```diff
@@ -0,0 +1,21 @@
+package api
+
+import (
+	"fmt"
+	"strings"
+)
+
+func (e *Error) Error() string {
```
danxmoran, line 8: Copy-paste from `influxdb` except for the referenced field names.

cmd/influx/main.go
```diff
@@ -1,26 +1,146 @@
...
+		if err := configs.Switch(configName); err != nil {
+			return config.Config{}, err
+		}
+	}
+	return configs.Active(), nil
+}
+
+// newApiClient returns an API client configured to communicate with a remote InfluxDB instance over HTTP.
+// Client parameters are pulled from the CLI context.
+func newApiClient(ctx *cli.Context, injectToken bool) (api.ClientInterface, error) {
```
danxmoran, line 50: I've manually tested that the `host`-related pieces of this work. It's hard to test that the auth-token things do what we want when testing against `/ping`; there will probably need to be follow-up fixes.

internal/config/config.go
```diff
@@ -0,0 +1,132 @@
+package config
```
danxmoran, line 1: This file is mostly copy-paste from InfluxDB, there are some methods that aren't used (yet). The unused pieces will be used as part of implementing `config` and `setup`

internal/config/local.go
```diff
@@ -0,0 +1,262 @@
+package config
```
danxmoran, line 1: I tried to simplify a few places in this code after copying it from `influxdb`, please call out anywhere else that could benefit from a refactor.

internal/ping/ping.go
```diff
@@ -0,0 +1,38 @@
...
+	}
+	parsed, err := api.ParseGetHealthResponse(resp)
+	if err != nil {
+		return fmt.Errorf("failed to parse health check response: %w", err)
+	}
+
+	if parsed.StatusCode()/100 != 2 {
+		return fmt.Errorf("health check failed (got status %d): %s", parsed.StatusCode(), string(parsed.Body))
+	}
+	if parsed.JSON200 == nil {
```
danxmoran, line 30: The codegen picks weird names for things, I think because we haven't set an `operationId` on the response schemas(?)

pkg/tracing/context.go
```diff
@@ -0,0 +1,28 @@
...
+
+import (
+	"context"
+
+	"github.com/influxdata/influx-cli/v2/api"
+)
+
+// Context is a convenience wrapper for tracking a global tracing
+// ID through CLI commands.
+type Context interface {
```
danxmoran, line 11: This was my thought for how to deal with the global trace ID parameter without adding an extra arg to every function in the program; open to other suggestions.

internal/cli.go
```diff
@@ -0,0 +1,15 @@
+package internal
+
+import (
+	"io"
+
+	"github.com/influxdata/influx-cli/v2/internal/api"
+)
+
+type CLI struct {
```
danxmoran, line 9: Decided I didn't like the custom `context` wrapper, made this as the replacement.

Self-review: needs a doc comment

internal/ping.go
```diff
@@ -0,0 +1,43 @@
+package internal
+
+import (
+	"context"
+	"fmt"
+
+	"github.com/influxdata/influx-cli/v2/internal/api"
+)
+
+type Client interface {
```
danxmoran, line 10:
```suggestion
type PingClient interface {
```

danxmoran: This pattern (defining a more focused interface instead of using the mega `api.ClientWithResponsesInterface`) is meant to help isolate unit test suites from one another. Adding a new route to the codegen ideally shouldn't require changes to unrelated tests.

dgnorton (approved):

PR #33, opened 2021-04-16
refactor: switch to OpenAPITools's generator for our HTTP clients

### LINKED ISSUE #32: Experiment with using OpenAPITools/openapi-generator
I picked [`deepmap/oapi-codegen`](https://github.com/deepmap/oapi-codegen) as the initial HTTP client generator used in this project because it seemed very lightweight and (mostly) idiomatic. I've run across one major incompatibility between its outputs and our preferred style of writing OpenAPI specs, and a few other rough edges.

[`OpenAPITools/openapi-generator`](https://github.com/OpenAPITools/openapi-generator) is a more heavyweight (but more actively supported) alternative for code generation. It looks like it may avoid the problem I hit with `deepmap`. Investigate what it'd take to swap it into the project, and cut over if it's not awful.

danxmoran:
Closes #32 

The `deepmap` generator was easy to run and lightweight, but its outputs didn't correctly handle use of the `servers` field to specify base paths in API routes. I still need to check that the `OpenAPITools` generator does the right thing there, but this is what the diff looks like to rewrite the existing code in `main`


go.sum
```diff
@@ -1,128 +1,395 @@
+cloud.google.com/go v0.26.0/go.mod h1:aQUYkXzVsufM+DwF1aE+0xfcU+56JwCaLick0ClmMTw=
```
danxmoran, line 1: I have no idea what all these GCP dependencies are for

etc/generate-openapi.sh
```diff
@@ -0,0 +1,66 @@
...
+  #
+  # See https://github.com/OpenAPITools/openapi-generator/issues/9280 for a feature request
+  # to make the Go generator's number-handling compliant with the spec, so we can generate int64
+  # fields directly without making our public docs invalid.
+  for m in internal/api/model_*.go; do
+    sed -i.bak -e 's/int32/int64/g' ${m}
+  done
+
+  rm internal/api/*.bak
+)
```
danxmoran, line 59: I am reluctant to add more awful hacks here, but I could probably add some logic to rename all the generated `X.go` files into `X.gen.go` to match our conventions in other projects.

internal/ping.go
```diff
@@ -7,37 +7,27 @@ import (
...
-	GetHealthWithResponse(context.Context, *api.GetHealthParams, ...api.RequestEditorFn) (*api.GetHealthResponse, error)
-}
-
 // Ping checks the health of a remote InfluxDB instance.
-func (c *CLI) Ping(ctx context.Context, client Client) error {
-	resp, err := client.GetHealthWithResponse(ctx, &api.GetHealthParams{ZapTraceSpan: c.TraceId})
+func (c *CLI) Ping(ctx context.Context, client api.HealthApi) error {
+	req := client.GetHealth(ctx)
+	if c.TraceId != "" {
+		req.ZapTraceSpan(c.TraceId)
```
danxmoran, line 14: This returns an `api.HealthApi`, but when I looked at the code I thought it might also mutate the original struct (yay Java builder pattern). Need to double-check.

etc/generate-openapi.sh
wbaker85: Do the generated files end up with a comment at the top that say that they were generated and not to edit them directly or something like that? If so, that'd be good enough for me.

danxmoran: They do, it looks like:
```
// Code generated by OpenAPI Generator (https://openapi-generator.tech); DO NOT EDIT.
```

wbaker85: I see that they definitely do, and that there is a nice `README` file in the folder with the generated files!

dgnorton (dismissed):

internal/ping.go
danxmoran: Good thing I double-checked!

cmd/influx/main.go
```diff
@@ -143,6 +136,10 @@ func main() {
 			Hidden:  true,
 			EnvVars: []string{"INFLUX_TRACE_DEBUG_ID"},
 		},
+		&cli.BoolFlag{
```
danxmoran, line 139: Useful to see what all this generated code ends up sending over the wire

wbaker85 (approved):

PR #34, opened 2021-04-19
refactor: replace bash hacks with custom templates

danxmoran:
The generator script introduced in #33 included some ugly hacks to delete/rewrite pieces of generated code that we don't like. I dug into OpenAPITools a bit more and found that it supports selectively overriding code templates, which feels like a much more robust approach.


wbaker85 (approved):
Looks good!

PR #35, opened 2021-04-20
feat: port `influx setup` from `influxdb`

### LINKED ISSUE #17: Port the `setup` command tree from `infludb`
Port the `setup` command (and any subcommands) from the `influxdb` repo.

danxmoran:
Closes #17 

Another big one:
1. Update `api.yml` to include the routes and schemas used by `influx setup`, and regenerate our client
2. Port helper functionality from `influxdb`
   * Duration-parsing
   * Tab-writing
   * Console UI
3. Port the `setup` command from `influxdb`
   * Refactor existing code so the main `CLI` struct references a local config service, so `setup` can write configs to disk
4. Write some tests

To manually test this:
1. `make influx` in this repo to build the CLI
2. `make influxd` in the `influxd` repo to build the server
3. `rm -r ~/.influxdbv2` to wipe any existing configs / metadata
4. Run `influxd`, wait for the HTTP "Listening" log
5. Run `influx setup`


cmd/influx/main.go
```diff
@@ -160,18 +174,102 @@ func main() {
 				Usage: "Check the InfluxDB /health endpoint",
 				Flags: commonFlags,
```
wbaker85, line 175: I don't think you need the auth token to hit the `/health` endpoint: https://github.com/influxdata/influxdb/blob/7bde3413b3801bdb15095ddb492e4ea2e916fc28/http/handler.go#L123

cmd/influx/main.go
```diff
@@ -160,18 +174,102 @@ func main() {
...
+					&cli.BoolFlag{
+						Name:    printJsonFlag,
+						Usage:   "Output data as JSON",
+						EnvVars: []string{"INFLUX_OUTPUT_JSON"},
+					},
+					&cli.BoolFlag{
+						Name:    hideHeadersFlag,
+						Usage:   "Hide the table headers in output data",
+						EnvVars: []string{"INFLUX_HIDE_HEADERS"},
+					},
```
wbaker85, line 245: Would it be useful to separate this list of flags out as a separate variable like `commonFlags` and `commonFlagsNoToken` are their own variables? Maybe not, since I'd imagine `commonFlags` and `commonFlagsNoToken` will be used in many other commands, but just a thought.

danxmoran: `printJsonFlag` and `hideHeadersFlag` can/should be factored out eventually, as they're common to many commands. The rest I'd like to keep as-is because `setup` has unique descriptions for each, and IMO it's nice to have minimal distance between where the flags are defined and where they're looked up in the `Action` below.

wbaker85 (approved):

PR #39, opened 2021-04-23
refactor: move mock APIs and services into mock module

danxmoran:
I made this cleanup when I first started working on #25 and thought there would be a lot of mocks needed for it. It turned out to not be the case, but I still think this is an improvement in the repo's organization.


wbaker85 (approved):
Looks good!

PR #40, opened 2021-04-23
fix: fix some issues with the local config service

danxmoran:
Unrelated fixes I found while working on #25


wbaker85 (approved):
:thumbsup:

PR #41, opened 2021-04-23
feat(api): improve codegen outputs

danxmoran:
Part of #25 

* Use `[]byte` for generated request bodies when the source schema has `format: byte`
* Gzip-compress request bodies when `Content-Encoding: gzip` is set
   * Add tests to show both gzip and no-gzip work
* Require that all models returned as error conditions in our API implement the `error` interface
   * Move the implementation for `api.HealthCheck` out of `ping.go` and into `api/error.go`

### DIFF
--- a/internal/api/client_internal_test.go
+++ b/internal/api/client_internal_test.go
@@ -12,7 +12,7 @@ import (
 
 func TestNoGzipRequest(t *testing.T) {
 	client := APIClient{cfg: NewConfiguration()}
-	body := []byte("This should get gzipped")
+	body := []byte("This should not get gzipped")
 	req, err := client.prepareRequest(
 		context.Background(),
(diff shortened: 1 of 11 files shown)


internal/api/client_internal_test.go
```diff
@@ -0,0 +1,52 @@
...
+	"context"
+	"io"
+	"testing"
+
+	"github.com/stretchr/testify/require"
+)
+
+func TestNoGzipRequest(t *testing.T) {
+	client := APIClient{cfg: NewConfiguration()}
+	body := []byte("This should get gzipped")
```
wbaker85, line 15:
```suggestion
	body := []byte("This should not get gzipped")
```

wbaker85 (approved):
Looks good!

PR #42, opened 2021-04-23
feat: port write-batcher from influxdb

danxmoran:
Part of #25 

This is pretty close to a verbatim copy of the `write` module from `influxdb`. Apart from some renamed structs, the main difference is:
* The batcher in `influxdb` was hard-coded to take an HTTP client and call the `/write` API per-batch
* The batcher here takes a generic `writeFn`

### DIFF
--- a/internal/batcher/buffer_batcher.go
+++ b/internal/batcher/buffer_batcher.go
@@ -1,0 +1,164 @@
+package batcher
+
+import (
+	"bufio"
+	"bytes"
+	"context"
+	"errors"
+	"io"
+	"time"
+)
+
+const (
+	// DefaultMaxBytes is 500KB; this is typically 250 to 500 lines.
+	DefaultMaxBytes = 500000
+	// DefaultInterval will flush every 10 seconds.
+	DefaultInterval = 10 * time.Second
+)
+
+var (
+	// ErrLineTooLong is the error returned when reading a line that exceeds MaxLineLength.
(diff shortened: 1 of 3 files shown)


internal/batcher/buffer_batcher.go
```diff
@@ -0,0 +1,164 @@
...
+	}
+	err := scanner.Err()
+	if errors.Is(err, bufio.ErrTooLong) {
+		err = ErrLineTooLong
+	}
+	errC <- err
+}
+
+// finishes when the lines channel is closed or context is done.
+// if an error occurs while writing data to the write service, the error is send in the
```
wbaker85, line 91:
```suggestion
// if an error occurs while writing data to the write service, the error is sent in the
```

internal/batcher/buffer_batcher_test.go
```diff
@@ -0,0 +1,192 @@
...
+			// mocking the batcher service here to either return an error
+			// or get back all the bytes from the reader.
+			var (
+				got        string
+				gotFlushes int
+			)
+			err := b.WriteBatches(context.Background(), tt.args.r(), func(batch []byte) error {
+				got = string(batch)
+				gotFlushes++
+				return nil
```
wbaker85, line 168: Looks like the `writeFn` used for this and the previous use of `b.WriteBatches` is the same, so you could consider extracting it to DRY things up if it makes sense.

wbaker85 (approved):
:+1:

PR #44, opened 2021-04-26
feat: port rate-limiter from influxdb

danxmoran:
Part of #25 

`ToBytesPerSecond` is copied verbatim from `influxdb`. The rest of the logic was factored out of the `createLineReader` function in `cmd/influx/write.go`, for better separation of concerns / easier testing.


wbaker85 (approved):
Looks good!

PR #47, opened 2021-04-26
feat: port influx write from influxdb

### LINKED ISSUE #25: Port the `write` command tree from `influxdb`
Port the `write` command (and any subcommands) from the `influxdb` repo.

danxmoran:
Closes #25 

I did some manual testing, and I think the unit tests have decent coverage. We need to add some robust e2e tests but I think that'll get easier once we port `influx query`.


wbaker85 (approved):
Looks good!

PR #50, opened 2021-04-27
refactor: inject trace ID as a default header

### LINKED ISSUE #49: Refactor how we set trace ID on HTTP requests
The CLI takes a (hidden) argument for specifying a trace ID to set on all HTTP requests made while executing commands. We've been passing this flag into every command and using codegen'd methods to set it on requests. This is repetitive, error-prone, and annoying to test.

Refactor to inject the trace ID into the "default" headers used by the HTTP client, in the same way we inject the auth token.

danxmoran:
Closes #49


wbaker85 (approved):
Looks good!

Issue #51, opened 2021-04-27
Add commands for managing explicit bucket schemas


danxmoran:
InfluxDB Cloud is adding support for explicit bucket schemas. We want to be able to manage those schemas via the CLI.

* Modify the bucket create command to allow specifying a schema-type of either implicit (default) or explicit
* Modify the bucket list command to show the schemaType property
* Add new commands to list, create and update existing measurement schemas for an explicit schema bucket

See influxdata/openapi#34 for the OpenAPI side of things

danxmoran:
Closed by #52

PR #52, opened 2021-04-28
feat: Add bucket schema management commands

stuartcarnie:
This PR extends the tool with commands to manage explicit schema buckets per #51.

Changes include:

* Modified the `bucket create` command to allow specifying a schema-type of either `implicit` (default) or `explicit`
* Modified the `bucket list` command to show the schemaType property as a new column, `Schema Type`
* Added a new `bucket-schema` command, with the following sub-commands:
    * `create` to create a new measurement schema for an existing `explicit` schema bucket;
    * `update` to add new columns to an existing measurement schema; and
    * `list` to find and print information about existing measurement schemas.


internal/cmd/measurement_schema/client.go
```diff
@@ -0,0 +1,241 @@
...
+		CreateMeasurementSchema(ctx, ids.BucketID).
+		OrgID(ids.OrgID).
+		MeasurementSchemaCreateRequest(api.MeasurementSchemaCreateRequest{
+			Name:    params.Name,
+			Columns: params.Columns,
+		}).
+		Execute()
+
+	if err != nil {
+		return err
```
danxmoran, line 97: Worth wrapping with a descriptive prefix?

internal/cmd/measurement_schema/client.go
```diff
@@ -0,0 +1,241 @@
...
+	res, _, err := c.BucketSchemasApi.
+		UpdateMeasurementSchema(ctx, ids.BucketID, id).
+		OrgID(ids.OrgID).
+		MeasurementSchemaUpdateRequest(api.MeasurementSchemaUpdateRequest{
+			Columns: params.Columns,
+		}).
+		Execute()
+
+	if err != nil {
+		return err
```
danxmoran, line 138: Worth wrapping?

internal/cmd/measurement_schema/client.go
```diff
@@ -0,0 +1,241 @@
...
+		GetMeasurementSchemas(ctx, ids.BucketID).
+		OrgID(ids.OrgID)
+
+	if params.Name != "" {
+		req = req.Name(params.Name)
+	}
+
+	res, _, err := req.Execute()
+	if err != nil {
+		return err
```
danxmoran, line 166: Worth wrapping?

internal/cmd/measurement_schema/client.go
```diff
@@ -0,0 +1,241 @@
...
+
+	req := c.BucketSchemasApi.
+		GetMeasurementSchemas(ctx, ids.BucketID).
+		OrgID(ids.OrgID)
+
+	if params.Name != "" {
+		req = req.Name(params.Name)
+	}
+
+	res, _, err := req.Execute()
```
danxmoran, line 164: I like this style of calling `Execute()` on the request, will make a refactor ticket to switch existing files over to it.

internal/cmd/measurement_schema/client.go
```diff
@@ -0,0 +1,241 @@
...
+		return nil
+	}
+
+	if c.CLI.PrintAsJSON {
+		return c.CLI.PrintJSON(m)
+	}
+
+	var headers []string
+	if extended {
+		headers = []string{
```
danxmoran, line 182: The existing commands in `influx` tend to use more human-readable column headers in table mode, for example: https://github.com/influxdata/influx-cli/pull/54/files#diff-425b59256130c1373f5df8900912229d394ed1ed4debba3814d7cc9c77d0ec40R277-R281

Should this command use the same style?

cmd/influx/measurement_schema.go
```diff
@@ -0,0 +1,247 @@
...
+	i, ok := ctx.App.Metadata["measurement_schema"].(measurement_schema.Client)
+	if !ok {
+		panic("missing measurement schema client")
+	}
+	return i
+}
+
+func newMeasurementSchemaCmd() *cli.Command {
+	return &cli.Command{
+		Name:  "bucket-schema",
```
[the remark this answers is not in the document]
stuartcarnie: I did try that but switch to a top-level command, as I think it is more discoverable and follows `tool <noun> <verb>` . 

I can't think of examples where it is `tool <noun> <noun> <verb>`, can you?

internal/bucket.go
```diff
@@ -48,6 +50,19 @@ func (c *CLI) BucketsCreate(ctx context.Context, clients *BucketsClients, params
 	if params.Description != "" {
 		reqBody.Description = &params.Description
 	}
+
+	if params.SchemaType != "" {
+		d, err := json.Marshal(params.SchemaType)
+		if err != nil {
+			return fmt.Errorf("unable to marshal schema type: %w", err)
+		}
+		var st api.SchemaType
```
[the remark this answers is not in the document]
stuartcarnie: I've switched to that

internal/cmd/measurement_schema/client.go
```diff
@@ -0,0 +1,241 @@
...
+		Execute()
+	if err != nil {
+		return "", err
+	}
+
+	if len(res.MeasurementSchemas) == 0 {
+		return "", fmt.Errorf("measurement schema not found %q", name)
+	}
+
+	return res.MeasurementSchemas[0].Id, nil
```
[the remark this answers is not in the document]
stuartcarnie: Not when calling with the `name` query parameter, as there is a unique index of names per bucket on the server.

internal/cmd/measurement_schema/client.go
```diff
@@ -0,0 +1,241 @@
...
+
+	if params.BucketID.Valid() {
+		req = c.BucketApi.GetBuckets(ctx).Id(params.BucketID.String())
+	} else if params.BucketName != "" {
+		req = c.BucketApi.GetBuckets(ctx).Name(params.BucketName)
+	} else {
+		return nil, errors.New("must specify bucket ID or bucket name")
+	}
+
+	resp, _, err := c.BucketApi.GetBucketsExecute(req)
```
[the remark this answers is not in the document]
stuartcarnie: I'll add a check to short circuit the call when the IDs are given.

internal/cmd/measurement_schema/client.go
stuartcarnie: Yes, and done 👍🏻

internal/cmd/measurement_schema/client.go
stuartcarnie: Done 👍🏻

cmd/influx/measurement_schema.go
```diff
@@ -0,0 +1,247 @@
...
+			)...,
+		),
+		Action: func(ctx *cli.Context) error {
+			var (
+				r    io.Reader
+				name string
+			)
+
+			arg := ctx.Args().First()
+			if arg == "-" || params.ColumnsFile == "" {
```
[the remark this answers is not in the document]
stuartcarnie: I've moved it to a shared function in the `measurement_schema` package.

internal/cmd/measurement_schema/client.go
```diff
@@ -0,0 +1,241 @@
+package measurement_schema
```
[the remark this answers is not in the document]
stuartcarnie: The benefit of `cmd/<command-name>` is everything under `cmd/*` should be the name of a subcommand. If more complex commands require helper modules, what do you do if / when that helper is used by more than one command? Do you move it as a sibling, breaking the consistency of dir = top-level command?

There is precedence for this layout, per:

* gh cli: https://github.com/cli/cli/tree/trunk/pkg/cmd
* kubectl cli: https://github.com/kubernetes/kubectl/tree/master/pkg/cmd

WDYT?

stuartcarnie:
NOTE: I've modified the script to reference the updated spec on my branch in the `openapi` to test CI. I've opened https://github.com/influxdata/openapi/pull/34 for review and will remove the last commit when it merges to `main`

cmd/influx/measurement_schema.go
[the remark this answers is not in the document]
danxmoran: In the current CLI we have `v1 auth`, `v1 dbrp`, `task log`, and `task run` ("run" as a noun, not a verb), but the dominant design is `<noun> <verb>`.

Ok by me to leave it as-is for now :+1: we can always change it later if product wants.

cmd/influx/measurement_schema.go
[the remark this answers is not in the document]
danxmoran: That's a fair point. My thinking wasn't so much to make `internal/` entirely abstracted from being a CLI (for example, everything is driven through the `CLI` struct), but instead to abstract it away from `urfave/cli` in case we ended up needing to switch back to cobra. Not sure if that's a useful enough separation to maintain, but it does allow for things like stdin detection to be unit-tested more easily.

internal/bucket_test.go
```diff
@@ -94,9 +94,9 @@ func TestBucketsCreate(t *testing.T) {
 		{
 			name: "retention but not shard-group duration",
 			params: internal.BucketsCreateParams{
```
danxmoran, line 96: Can you add tests for:
* Create w/ `explicit`
* List w/ combo of `implicit` and `explicit`?

Instead of new tests, I think it'd be ok to modify some of the existing ones.

internal/cmd/measurement_schema/client.go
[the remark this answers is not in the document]
danxmoran: That makes sense to me :+1:

internal/cmd/measurement_schema/client.go
danxmoran: I see the human-friendly headers in the helper functions, but it looks like these slices still need updated to match

internal/cmd/measurement_schema/client.go
```diff
@@ -0,0 +1,284 @@
...
+	BucketSchemasApi api.BucketSchemasApi
+	CLI              *internal.CLI
+}
+
+type orgBucketID struct {
+	OrgID    string
+	BucketID string
+}
+
+func (c Client) resolveMeasurement(ctx context.Context, ids orgBucketID, name string) (string, error) {
```
danxmoran, line 26: Not sure if it matters, but should these methods take pointer receivers?

internal/cmd/measurement_schema/client.go
```diff
@@ -0,0 +1,284 @@
...
+type CreateParams struct {
+	internal.OrgBucketParams
+	Name           string
+	Stdin          io.Reader
+	ColumnsFile    string
+	ColumnsFormat  ColumnsFormat
+	ExtendedOutput bool
+}
+
+func (c Client) Create(ctx context.Context, params CreateParams) error {
```
danxmoran, line 112: Maybe overkill, but when thinking about file layout I'd considered splitting up the main entry-points for subcommands into their own files, i.e. having a `create.go`, `update.go`, etc. WDYT?

internal/cmd/measurement_schema/client.go
```diff
@@ -0,0 +1,284 @@
...
+	"errors"
+	"fmt"
+	"io"
+	"os"
+
+	"github.com/influxdata/influx-cli/v2/internal"
+	"github.com/influxdata/influx-cli/v2/internal/api"
+)
+
+type Client struct {
```
danxmoran, line 15: WDYT about writing unit tests for the logic in this file? It felt tedious when I did it for the `bucket` commands, but it flushed out some bugs in the process.

danxmoran (commented):
This can rebase against `main`

etc/generate-openapi.sh
```diff
@@ -4,13 +4,15 @@ declare -r ETC_DIR="$(cd "$(dirname "${BASH_SOURCE[0]}")" &> /dev/null && pwd)"
 declare -r ROOT_DIR="$(dirname ${ETC_DIR})"
 
 declare -r GENERATOR_DOCKER_IMG=openapitools/openapi-generator-cli:v5.1.0
-declare -r OPENAPI_COMMIT=e41d5e327a67e472a46cd6edfe673496e1b066dd
+declare -r OPENAPI_COMMIT=sgc/issues/9672
```
danxmoran, line 7: When you reset this and re-generate, the changes from #57 should take effect. I expect the only impact will be that the new files added here will get renamed to `*.gen.go`

cmd/influx/measurement_schema.go
```diff
@@ -0,0 +1,189 @@
...
+				&cli.StringFlag{
+					Name:        "columns-file",
+					Usage:       "A path referring to list of column definitions",
+					Destination: &params.ColumnsFile,
+				},
+				&cli.GenericFlag{
+					Name:        "columns-format",
+					Usage:       "The format of the columns file. \"auto\" will attempt to guess the format.",
+					DefaultText: "auto",
+					Value:       &params.ColumnsFormat,
```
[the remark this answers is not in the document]
stuartcarnie: Interesting – I guess it doesn't, as the demo worked… I will change to `Destination` for consistency.

cmd/influx/bucket.go
```diff
@@ -91,6 +94,12 @@ func newBucketCreateCmd() *cli.Command {
 				EnvVars:     []string{"INFLUX_ORG"},
 				Destination: &params.OrgName,
 			},
+			&cli.GenericFlag{
+				Name:        "schema-type",
+				Usage:       "The schema type (implicit, explicit)",
+				DefaultText: "implicit",
+				Value:       &params.SchemaType,
+			},
```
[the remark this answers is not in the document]
stuartcarnie: No these are `GenericFlag` types, which only have a `Value`

cmd/influx/measurement_schema.go
```diff
@@ -0,0 +1,189 @@
...
```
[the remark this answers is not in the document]
stuartcarnie: No these are `GenericFlag` types, which only have a `Value`

cmd/influx/measurement_schema.go
```diff
@@ -0,0 +1,189 @@
...
+		Usage:  "Update a measurement schema for a bucket",
+		Before: withMSClient(),
+		Flags: append(
+			commonFlags,
+			append(
+				getOrgBucketFlags(&params.OrgBucketParams),
+				&cli.GenericFlag{
+					Name:  "id",
+					Usage: "ID of the measurement",
+					Value: &params.ID,
```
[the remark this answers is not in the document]
stuartcarnie: No these are `GenericFlag` types, which only have a `Value`

etc/generate-openapi.sh
```diff
@@ -4,13 +4,15 @@ declare -r ETC_DIR="$(cd "$(dirname "${BASH_SOURCE[0]}")" &> /dev/null && pwd)"
...
 
 # Download our target API spec.
 # NOTE: openapi-generator supports HTTP references to API docs, but using that feature
 # causes the host of the URL to be injected into the base paths of generated code.
 curl -o ${ROOT_DIR}/internal/api/cli.yml https://raw.githubusercontent.com/influxdata/openapi/${OPENAPI_COMMIT}/contracts/cli.yml
 
+# cp -c ~/projects/go/influxdata/openapi/contracts/cli.yml ${ROOT_DIR}/internal/api/cli.yml
+
```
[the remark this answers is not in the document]
stuartcarnie: Done 👍🏻

stuartcarnie:
> This can rebase against `main`

Not sure what happened, but however `sgc/experimental` was merged to `main` caused the branches to diverge, so I had to cherry pick all the individual commits. 🤷🏻

internal/cmd/measurement_schema/client.go
stuartcarnie: I'll add some today 👍🏻

internal/cmd/measurement_schema/client.go
stuartcarnie: I've moved to using constants to reduce the chance of error

internal/bucket_test.go
stuartcarnie: Done

cmd/influx/measurement_schema.go
```diff
@@ -0,0 +1,247 @@
...
+			)...,
+		),
+		Action: func(ctx *cli.Context) error {
+			var (
+				r    io.Reader
+				name string
+			)
+
+			arg := ctx.Args().First()
+			if arg == "-" || params.ColumnsFile == "" {
```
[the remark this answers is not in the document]
stuartcarnie: 👍🏻 I've moved it

cmd/influx/params.go
```diff
@@ -0,0 +1,36 @@
+package main
+
+import (
+	"github.com/influxdata/influx-cli/v2/internal"
+	"github.com/urfave/cli/v2"
+)
+
```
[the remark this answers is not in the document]
stuartcarnie: I think you could move the variables to this package and leave the bindings in `main` – that keeps with your mantra of leaving `urfave` logic in `cmd`

internal/cmd/measurement_schema/client.go
stuartcarnie: I've kept it as a value receiver to indicate that `Client` is immutable. `Client` ultimately serves as a namespace for the functions.

internal/cmd/measurement_schema/client.go
stuartcarnie: I don't think it is necessary at this point, however, if the commands (`List`, `Create`, etc) were to have their own distinct state and inputs, it might make sense to separate them. I'll rename the package to `bucket_schema` to match the sub-command name.

internal/cmd/measurement_schema/client.go
stuartcarnie: Done.

I introduced the Go team's mock package, https://github.com/golang/mock, as we've also been using it in IDPE and it's been really useful.

stuartcarnie:
Hi @danxmoran – let me know if there are still any outstanding questions you have. I've added a number of unit tests for all the `bucket-schema` commands.

danxmoran (approved):
Thanks for the contribution @stuartcarnie! Exciting to see a cloud-focused feature land early.

PR #53, opened 2021-04-29
refactor: Move from global state to functions

stuartcarnie:
This PR represents a few experiments of features I've used in Cobra and some attempts to reduce duplication of code.

1. Uses `cli.GenericFlag` to encapsulate parsing and validation of flag values at parse time. This removes the burden from the individual CLI commands to parse and validate args and options.

2. Add `flag.ID` that may be used by any flag that requires an Influx ID. `flag.ID` parses and validates string value is a valid ID, removing this burden from individual commands and ensuring valid values before the command actions begins.

3. Adds a `Flags()` method to the Params structs to directly capture the values when parsing flags.

4. Moves from global state to local builder functions for the majority of the commands. This allows the commands to bind to flag variables reducing the repeated `ctx.String()`, `ctx.Int()`, etc

5. Leverages the `BeforeFunc` to create middleware and inject the CLI and API client into commands, saving the repeated boilerplate across all of the instantiated commands. This is extensible, so additional middleware can be appends using the `middleware.WithBeforeFns`


cmd/influx/bucket.go
```diff
@@ -2,223 +2,101 @@ package main
...
+	return middleware.WithBeforeFns(
+		withCli(),
+		withApi(),
+		func(ctx *cli.Context) error {
+			client := getAPI(ctx)
+			ctx.App.Metadata["bucketsClient"] = internal.BucketsClients{
+				BucketApi: client.BucketsApi,
+				OrgApi:    client.OrganizationsApi,
+			}
+			return nil
```
stuartcarnie, line 19: This creates a `cli.BeforeFunc` that ensures all the dependencies are available for bucket commands

cmd/influx/bucket.go
```diff
@@ -2,223 +2,101 @@ package main
...
-					OrgName: ctx.String("org"),
-				})
-			},
+	)
+}
+
+func getBucketsClient(ctx *cli.Context) internal.BucketsClients {
+	i, ok := ctx.App.Metadata["bucketsClient"].(internal.BucketsClients)
+	if !ok {
+		panic("missing buckets client")
```
stuartcarnie, line 27: we panic here, because it represents a programmer's error if this fails

cmd/influx/bucket.go
```diff
@@ -2,223 +2,101 @@ package main
...
+	}
+
+}
+
+func newBucketCreateCmd() *cli.Command {
+	var params internal.BucketsCreateParams
+	return &cli.Command{
+		Name:   "create",
+		Usage:  "Create bucket",
+		Before: withBucketsClient(),
```
stuartcarnie, line 51: We inject the dependencies for a command like this

cmd/influx/bucket.go
```diff
@@ -2,223 +2,101 @@ package main
...
+		Flags:  append(commonFlags, params.Flags()...),
+		Action: func(ctx *cli.Context) error {
+			clients := getBucketsClient(ctx)
+			return getCLI(ctx).BucketsCreate(ctx.Context, &clients, &params)
```
stuartcarnie, line 55: Extracting the dependencies using `getBucketsClient` and `getCLI`

cmd/influx/main.go
```diff
@@ -185,26 +186,72 @@ func newApiClient(ctx *cli.Context, cli *internal.CLI, injectToken bool) (*api.A
...
+	if !ok {
+		panic("missing APIClient without token")
+	}
+	return i
+}
+
 func main() {
-	if err := app.Run(os.Args); err != nil {
+	ctx := signals.WithStandardSignals(context.Background())
+	if err := app.RunContext(ctx, os.Args); err != nil {
```
stuartcarnie, line 254: We run the app with a context that contains the standard signals, saving us from having to do that for every sub command

cmd/influx/main.go
```diff
@@ -185,26 +186,72 @@ func newApiClient(ctx *cli.Context, cli *internal.CLI, injectToken bool) (*api.A
 	return api.NewAPIClient(apiConfig), nil
 }
 
-// standardCtx returns a context that will cancel on SIGINT and SIGTERM.
-func standardCtx(ctx *cli.Context) context.Context {
-	return signals.WithStandardSignals(ctx.Context)
-}
-
```
stuartcarnie, line 192: This is eliminated as we use the `cli.App#RunContext` API with a context already combined with standard signals.

internal/bucket.go
```diff
@@ -26,6 +27,51 @@ type BucketsCreateParams struct {
...
 }
 
+func (p *BucketsCreateParams) Flags() []cli.Flag {
+	return []cli.Flag{
+		&cli.StringFlag{
+			Name:        "name",
+			Usage:       "New bucket name",
+			Aliases:     []string{"n"},
+			EnvVars:     []string{"INFLUX_BUCKET_NAME"},
+			Destination: &p.Name,
```
stuartcarnie, line 37: We bind the values to the params structs, so we don't have to call `ctx.String`, etc

internal/linereader/multi_reader.go
```diff
@@ -29,6 +29,33 @@ const (
...
+		*i = InputFormatDerived
+	case "lp":
+		*i = InputFormatLP
+	case "csv":
+		*i = InputFormatCSV
+	default:
+		return fmt.Errorf("unsupported format: %q", v)
+	}
+	return nil
+}
```
stuartcarnie, line 44: We implement `cli.Generic` and then use the `cli.GenericFlag` flag to bind to these strongly-typed flags. These flags parse and validate the values before calling our individual command action functions.

cmd/influx/bucket.go
```diff
@@ -2,223 +2,101 @@ package main
...
+func newBucketDeleteCmd() *cli.Command {
+	var params internal.BucketsDeleteParams
+	return &cli.Command{
+		Name:   "delete",
+		Usage:  "Delete bucket",
+		Before: withBucketsClient(),
+		Flags:  append(commonFlags, params.Flags()...),
+		Action: func(ctx *cli.Context) error {
+			clients := getBucketsClient(ctx)
+			return getCLI(ctx).BucketsDelete(ctx.Context, &clients, &params)
```
stuartcarnie, line 69: All the bucket actions are reduced to 1-2 lines of code.

cmd/influx/main.go
```diff
@@ -185,26 +186,72 @@ func newApiClient(ctx *cli.Context, cli *internal.CLI, injectToken bool) (*api.A
...
+
+func getCLI(ctx *cli.Context) *internal.CLI {
+	i, ok := ctx.App.Metadata["cli"].(*internal.CLI)
+	if !ok {
+		panic("missing CLI")
+	}
+	return i
+}
+
+func withApi() cli.BeforeFunc {
```
danxmoran, line 221: Could we have this take a `bool` arg and only build/register one client, so we can drop the `getAPINoToken` function? Or is the extra explicitness of `getAPI` vs. `getAPINoToken` worth it?

internal/bucket.go
danxmoran: I liked having the `cli.Command` and `cli.Flag` definitions all under `cmd/` and not in `internal/`, could we refactor back to that?

:+1: to setting `Destination` instead of repeating the flag names in `ctx.String` etc. calls

internal/throttler/throttler.go
```diff
@@ -16,6 +16,10 @@ type Throttler struct {
 	bytesPerSecond float64
 }
 
+func NewThrottlerBytesPerSecond(bytesPerSec float64) *Throttler {
+	return &Throttler{bytesPerSecond: bytesPerSec}
+}
+
 func NewThrottler(rate string) (*Throttler, error) {
```
danxmoran, line 23: Do we need this func that takes a `string` now that we're parsing the bytes-per-sec at the "edge" of the CLI?

internal/write.go
```diff
@@ -36,6 +38,44 @@ type WriteParams struct {
 	Precision  api.WritePrecision
 }
 
+func (p *WriteParams) Flags() []cli.Flag {
```
danxmoran, line 41: Especially here, I'd prefer to have all the `cli.Flag` definitions in once place instead of splitting them between `cmd/` and `internal/`

internal/throttler/throttler.go
```diff
@@ -16,6 +16,10 @@ type Throttler struct {
 	bytesPerSecond float64
 }
 
+func NewThrottlerBytesPerSecond(bytesPerSec float64) *Throttler {
```
danxmoran, line 19: Should this be:
```suggestion
func NewThrottlerBytesPerSecond(bytesPerSec BytesPerSec) *Throttler {
```
for clarity?

pkg/influxid/id.go
```diff
@@ -0,0 +1,123 @@
...
+	ErrInvalidID = errors.New("invalid ID")
+
+	// ErrInvalidIDLength is returned when an ID has the incorrect number of bytes.
+	ErrInvalidIDLength = errors.New("id must have a length of 16 bytes")
+)
+
+// ID is a unique identifier.
+//
+// Its zero value is not a valid ID.
+type ID uint64
```
danxmoran, line 24: Higher-level question: does the CLI need to care about the difference between and ID and an arbitrary `string`? I was thinking about it while porting `influx bucket`, and had these assumptions:
* The server-side will always have its own validation for IDs
* The CLI will only ever be passing string IDs from CLI args -> server and string IDs from server -> stdout

Is it better for the CLI to be agnostic to what makes a "valid" ID as a future-proofing mechanism? Or better to push the error message as close to the original data entry as possible?

pkg/cmd/flag/bytes_per_seconds.go
```diff
@@ -0,0 +1,22 @@
+package flag
```
danxmoran, line 1: Naming nitpick: should the module be `pkg/cli` instead of `pkg/cmd`?

stuartcarnie: Hah, I originally had it under cli and moved it to cmd. I do agree, cli is better

cmd/influx/main.go
stuartcarnie: I originally had it that way - and then thought it looked clearer at the call site that you were requesting the API sans token. Don’t feel very strongly about either approach

internal/bucket.go
stuartcarnie: I don’t see a problem doing it that way - could just move it to a helper function that binds the flags to a pointer of the params struct. One function for each. 

Bear in mind that your `main` package is going to get quite large if you throw every registered command in there. Most complex CLI applications end up with packages for each command and even sometimes one for each sub-command.

danxmoran: Agreed about `main` getting big, and moving towards more sub-packages. I had the same thought around `internal`.

pkg/cmd/flag/bytes_per_seconds.go
stuartcarnie: Done

stuartcarnie: In fact I introduced a `throttler.BytesPerSec` type to move all the logic into the `throttler` package

internal/throttler/throttler.go
stuartcarnie: Done

internal/write.go
stuartcarnie: Done – I think the improvements are 👍🏻

internal/throttler/throttler.go
stuartcarnie: No, so I've removed it 👍🏻

cmd/influx/main.go
danxmoran: I also like the explicitness at the call site, the more that I look at it. It's probably fine in practice, but bothers me to construct 2 clients when you're guaranteed to only use 1.

Assuming we keep both `getAPI` and `getAPINoToken`, what do you think between these options?
1. Have `withApi` take a bool arg that determines whether or not a token is injected
   * The key used in `ctx.App.Metadata` is also determined by the bool
   * Users need to call the correct `get` function based on the bool they passed, otherwise they'll see a panic
2. Have separate `withApi` and `withApiNoToken` middlewares
   * Could be `var`s that call an extracted `makeFn`, to avoid duplication
   * Users need to call the correct `get` function based on the middleware they used, otherwise they'll see a panic
3. Leave things as-is, even though we're building a redundant client it's worth reducing the chances of a panic when devs copy-paste-adjust for new commands

cmd/influx/write.go
```diff
@@ -11,271 +11,237 @@ import (
...
-					Compression:                compression,
-					Encoding:                   ctx.String("encoding"),
-					Headers:                    ctx.StringSlice("header"),
-					SkipRowOnError:             ctx.Bool("skipRowOnError"),
-					SkipHeader:                 ctx.Int("skipHeader"),
-					IgnoreDataTypeInColumnName: ctx.Bool("xIgnoreDataTypeInColumnName"),
-					Debug:                      ctx.Bool("debug"),
-				}
+func (p *writeParams) makeErrorFile(ctx *cli.Context) (*os.File, error) {
+	if !ctx.IsSet("errors-file") {
```
danxmoran, line 60: The errors-file flag is wired up to a struct field, so could we change this to:
```suggestion
func (p *writeParams) makeErrorFile() (*os.File, error) {
	if p.ErrorsFile == "" {
```

pkg/cli/middleware/middleware.go
```diff
@@ -0,0 +1,27 @@
+package middleware
+
+import (
+	"github.com/urfave/cli/v2"
+)
+
+func WithBeforeFns(fns ...cli.BeforeFunc) cli.BeforeFunc {
```
danxmoran, line 7: Can you add godoc to these two functions describing the motivation?

pkg/cli/flag/write_precision.go
```diff
@@ -0,0 +1,33 @@
+package flag
+
+import (
+	"fmt"
+
+	"github.com/influxdata/influx-cli/v2/internal/api"
+)
+
+type WritePrecision api.WritePrecision
```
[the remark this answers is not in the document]
stuartcarnie: Interesting – as I did try to find a non-generated file, and must have missed that one! I'd prefer it to live closer to the type also. I'll move it there.

cmd/influx/main.go
stuartcarnie: Given how light weight the client is, I was't concerned about building a second one, however, I can just move back to the `withApi(bool)` and move some of the burden back to the developer to define the correct version for their CLI command.

pkg/influxid/id.go
stuartcarnie: I would advocate it is better to do some parsing if the format is known, to fail faster. No different to parsing an integer flag into an `int` arg, even if it is sent as a string to the server. If our ID format changes, we can amend the ID parser.

pkg/cli/flag/write_precision.go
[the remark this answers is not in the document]
stuartcarnie: Done

pkg/cli/middleware/middleware.go
stuartcarnie: Done

stuartcarnie:
Thanks for the review, @danxmoran. I've pushed a commit with the feedback suggestions.

cmd/influx/main.go
danxmoran: I went ahead and pushed the change to unblock merge

danxmoran:
Sorry to dirty the commit history, didn't want to rebase/force-push and disrupt your other open PR 😬

danxmoran (approved):
I'll merge when CI is green to unblock other work in the repo using the new middleware. Thanks for taking the time to contribute the improvements!

PR #58, opened 2021-05-03
fix: check server status before local config to improve error message

danxmoran:
Before this change, if a user tried to double-setup their local server they'd see `config name is required if you already have existing configs`. That's technically correct, but IMO the more useful error to see is `instance has already been set up` (it's also the error they'd see in the old CLI).

Rearrange the validation performed by `setup` to make the server-side check run first.


wbaker85:
:+1:

wbaker85 (approved):

PR #59, opened 2021-05-03
feat: add bash/zsh/powershell completions

### LINKED ISSUE #8: Port the `completion` command from `influxdb`
Port the `completion` command (and any subcommands) from the `influxdb` repo.

danxmoran:
Closes #8 

The `completions` command follows the CLI from `influxdb`, but the implementation is entirely different from `influxdb` because of the difference in underlying CLI library.

I've manually tested all three shells and confirmed this works.


wbaker85:
Comment/question: When you do the `source <(influx completion [whatever])`, is the auto-completion supposed to persist between shell sessions? I only tried it with `zsh` and it didn't seem to persist when I opened a new shell, but maybe it's just something on my end. Thinking it might be useful to include something about the autocompletion not being persistent (if it's actually not) somewhere, maybe to the `README` or the docs/both.

danxmoran:
It's not expected to persist between sessions; I'll expand the notes in the README to call that out & suggest adding to `~/.profile` etc

wbaker85 (approved):
👍

PR #60, opened 2021-05-04
refactor: improve codegen to support `influx query` rewrite

danxmoran:
Part of #14

It's hard to see the full picture without adding the `/query` endpoint, but these changes include:
1. When `Content-Encoding: gzip` is set on a response, gunzip the request body before attempting any processing
2. When a request has `format: binary` in its top-level response schema, bypass the logic that pulls the full response into memory
   * Instead return the `io.ReadCloser` from the `http.Response` (or the gunzip-wrapped body) directly
3. Remove the `*http.Response` from the set of return values generated for every operation
   * Prevents a confusing "which do I use?" situation where we return both the HTTP response body as an `io.ReadCloser` and the `*http.Response` containing a reference to the same body
   * Cleans up a lot of code, because we were never using the raw return value but had to include it in mocks etc. to match the interface

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -30,7 +30,11 @@ openapi:
 	./etc/generate-openapi.sh
 
 fmt: $(SOURCES_NO_VENDOR)
+	# Format everything, but the import-format doesn't match our desired pattern.
 	gofmt -w -s $^
+	# Remove unused imports.
+	go run golang.org/x/tools/cmd/goimports -w $^
+	# Format imports.
 	go run github.com/daixiang0/gci -w $^
 
 bin/$(GOOS)/influx: $(SOURCES)
--- a/internal/api/templates/api.mustache
+++ b/internal/api/templates/api.mustache
@@ -38,7 +38,7 @@ type {{classname}} interface {
 	 * {{nickname}}Execute executes the request{{#returnType}}
 	 * @return {{{.}}}{{/returnType}}
 	 */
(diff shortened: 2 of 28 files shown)


Makefile
```diff
@@ -30,7 +30,11 @@ openapi:
 	./etc/generate-openapi.sh
 
 fmt: $(SOURCES_NO_VENDOR)
+	# Format everything, but the import-format doesn't match our desired pattern.
 	gofmt -w -s $^
+	# Remove unused imports.
+	go run golang.org/x/tools/cmd/goimports -w $^
```
danxmoran, line 36: This is a kludge to work around a bad piece of codegen.

The default codegen for `format: binary` pulls the full response body into memory and massages it into a `*os.File`. I was able to change the template to skip all that and return a direct `io.ReadCloser`, but I wasn't able to prevent the now-unneeded `import "os"` from appearing without breaking imports in other generated files.

`goimports` seems to be the only tool that will detect & remove unused imports. Luckily it seems to run quickly.

internal/api/templates/api.mustache
```diff
@@ -4,7 +4,9 @@ package {{packageName}}
 {{#operations}}
 import (
 	"bytes"
+    _gzip "compress/gzip"
```
danxmoran, line 7: Self review: make sure all whitespace here gets converted to tabs

wbaker85 (approved):
Looks good to me!

PR #65, opened 2021-05-07
refactor: update codegen to use schema-less flux AST

danxmoran:
The API spec for our Flux AST is complex enough that its codegen'd client code is very unwieldy to use. Since we only ever use a hardcoded subset of the AST, move to a schemaless model and accept the `map[string]interface{}` usage.


etc/generate-openapi.sh
```diff
@@ -6,7 +6,7 @@ declare -r API_DIR="${ROOT_DIR}/internal/api"
 
 declare -r GENERATED_PATTERN='^// Code generated .* DO NOT EDIT\.$'
 declare -r GENERATOR_DOCKER_IMG=openapitools/openapi-generator-cli:v5.1.0
-declare -r OPENAPI_COMMIT=4f62aa6175c6f8e9a425ea8e3e143c8c93f2c652
+declare -r OPENAPI_COMMIT=1e07d10aa1ec5c9e06fd4fa1d8dcce993d1d4bd8
```
danxmoran, line 9: Needs to be updated again after influxdata/openapi#46 merges

wbaker85 (approved):

PR #86, opened 2021-05-12
feat: port most of `influx query` from influxdb

danxmoran:
Part of #14 

Only raw-output mode is supported at the moment. Formatted output will come in a follow-up PR, still working through the edge-cases.


cmd/influx/query.go
```diff
@@ -0,0 +1,138 @@
...
+		}
+		return string(queryBytes), err
+	}
+
+	if file != "" {
+		return readFile(file)
+	}
+	if nargs == 0 {
+		return readStdin()
+	}
```
wbaker85, line 127: I'm wondering if there's a way to combine some of the logic from line 106 through 127. Could you have a single function that takes an `io.Reader` (or even an `*os.File`) instead of the separate `readFile` and `readStdin` functions? Then if you've got a file, open it with `os.Open` first and pass it into the function.

internal/cmd/query/query.go
```diff
@@ -0,0 +1,151 @@
...
+// See the docs for more info: https://docs.influxdata.com/influxdb/cloud/reference/flux/stdlib/profiler/
+func BuildExternAST(profilers []string) *api.Extern {
+	// Construct AST statements to import and set the 'profilers' option.
+	// NOTE: We've purposefully codegen'd a map[string]interface{} schema
+	// for the field populated here because the API spec for our Flux AST
+	// generates very hard-to-use models, and our attempts to change
+	// that have all (so far) broken the codegen for the UI.
+	//
+	// We assume that this logic will be changed infrequently enough that
+	// the lack of type-safety won't be frequent pain point.
```
wbaker85, line 67:
```suggestion
	// the lack of type-safety won't be a frequent pain point.
```

wbaker85, line 67:
```suggestion
	// the lack of type-safety won't be a frequent pain point.
```

cmd/influx/query.go
danxmoran: I originally made a single function that took an `io.Reader`, but it just pushed the redundancies around: each of the current `readFile` calls were replaced by something like:
```go
f, err := os.Open(path)
if err != nil {
  return err
}
defer f.Close()
query, err := readQuery(f, path)
if err != nil {
  return err
}
```
Making separate functions for read-path and read-stdin at least lets us use different `ioutil` conveniences and cut down the boilerplate.

wbaker85 (approved):
lgtm!

PR #87, opened 2021-05-12
feat: add formatted output to `influx query`

### LINKED ISSUE #14: Port the `query` command tree from `influxdb`
Port the `query` command (and any subcommands) from the `influxdb` repo.

danxmoran:
Closes #14 

The implementation here is a blend of existing code from other projects & my adaptations.
* `pkg/fluxcsv` is an adaptation of the query-result data structures & parsers from [`influxdata/influxdb-client-go`](https://github.com/influxdata/influxdb-client-go/)
  * I adjusted the logic to properly parse where new query results/tables start, instead of assuming a new table starts any time there's a new set of annotation rows
* `formatting_printer.go` is an adaptation of [this code](https://github.com/influxdata/influxdb/blob/master/cmd/influx/query.go#L216-L493) from `influxdb`, which itself was copy-pasted from [this file](https://github.com/influxdata/flux/blob/master/execute/format.go) in `flux`.
  * Apart from adapting to the new types in `pkg/fluxcsv`, I tried to simplify some of the data structures used to track parsing state.

While merging the two pieces of code, I noticed that the table-formatter in `influxdb` doesn't cover all the data types mentioned by our Go client. I need to track down the right behavior for those types.

### DIFF
--- a/internal/cmd/query/formatting_printer.go
+++ b/internal/cmd/query/formatting_printer.go
@@ -21,6 +21,7 @@ type formattingPrinter struct {
 	fmtBuf [64]byte
 
 	cols []fluxcsv.FluxColumn
+	lastColIdx int
 }
 
 
@@ -72,6 +73,7 @@ func (f *formattingPrinter) write(res *fluxcsv.QueryTableResult, out io.Writer) 
 		if res.AnnotationsChanged() {
 			// Reset and sort cols
 			f.cols = res.Metadata().Columns()
+			f.lastColIdx = len(f.cols) - 1
 			groupKeys := make(map[string]int, len(res.Metadata().GroupKeyCols()))
 			for i, k := range res.Metadata().GroupKeyCols() {
 				groupKeys[k] = i
@@ -112,7 +114,7 @@ func (f *formattingPrinter) write(res *fluxcsv.QueryTableResult, out io.Writer) 
 			w.write(eol)
 		}
 		if res.TableIdChanged() || res.AnnotationsChanged() {
-			w.write([]byte("TableId: keys: ["))
+			w.write([]byte("Table: keys: ["))
 			labels := make([]string, len(res.Metadata().GroupKeyCols()))
 			for i, c := range res.Metadata().GroupKeyCols() {
 				labels[i] = c
@@ -158,7 +160,9 @@ func (f *formattingPrinter) write(res *fluxcsv.QueryTableResult, out io.Writer) 
 				w.write(buf[:f.widths[i]-3])
 				w.write([]byte{'.', '.', '.'})
 			}
-			w.write(f.pad[:2])
+			if i != f.lastColIdx {
+				w.write(f.pad[:2])
+			}
 			if l > f.newWidths[i] {
 				f.newWidths[i] = l
 			}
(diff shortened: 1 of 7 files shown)


internal/cmd/query/formatting_printer.go
```diff
@@ -0,0 +1,269 @@
...
+	_, err := w.w.Write(data)
+	w.err = err
+}
+
+var minWidthsByType = map[fluxcsv.ColType]int{
+	fluxcsv.BoolDatatype:         12,
+	fluxcsv.LongDatatype:         26,
+	fluxcsv.ULongDatatype:        27,
+	fluxcsv.DoubleDatatype:       28,
+	fluxcsv.StringDatatype:       22,
```
wbaker85, line 60: General comment/question...I know these min widths are consistent with the existing CLI but some of them seem a little excessive. Like for a string, it seems like it ends up with a lot of wasted space for many strings. I'm not necessarily suggesting you change the formatting here, but I am curious if these min widths have some significance outside of appearance?

wbaker85: I could image some ideal future where the column min widths are based on the actual terminal size of the user, but that's probably a battle for another day :smile:

pkg/fluxcsv/query_result.go
```diff
@@ -0,0 +1,297 @@
...
+	if q.err == io.EOF {
+		q.err = nil
+		return false
+	}
+	if q.err != nil {
+		return false
+	}
+
+	if len(row) <= 1 {
+		goto readRow
```
wbaker85, line 151: Feels like I'd be remiss if I didn't make a comment about using `goto`...looks like it was brought over from `influxdata/influxdb-client-go` and it seems to work out fine here so I'm good with it. Might get ugly with a bunch of `break` and `continue`.

wbaker85:
Something I noticed when playing around with this CLI at this point to look at the output: When I just type `influx query` with no arguments or flags, my terminal hangs and I can't get it back by killing it with `ctrl-c`. When using the CLI from `influxdb`, doing this will cause the terminal to appear to hang, but I can kill it with `ctrl-c` and get the terminal back. Suspending or EOF'ing either one works fine though.

wbaker85 (approved):
Looks good! Nothing blocking, just made a couple comments for consideration.

internal/cmd/query/formatting_printer.go
danxmoran: As far as I can tell the only min width that makes sense here is the one for time columns. Detecting based on terminal width does sound very tempting... it's very annoying to use `influx query` as-is when I've got it on my vertical monitor.

I'll make a ticket!

pkg/fluxcsv/query_result.go
danxmoran: I'll stare at it hard before merging to see how painful it'd be to remove.

danxmoran: Hmmm, so I think it'd be possible to refactor into a `for{}` and use `continue` where all the `goto`s are now, but we'd still need to label the `for` so we could `break` it from inside of the `switch` statement. Not sure if it's worth adding the extra layer of indentation, I think `goto` is in Go to handle situations like this.

PR #90, opened 2021-05-13
feat: port `influx config` from `influxdb`

### LINKED ISSUE #9: Port the `config` command tree from `influxdb`
Port the `config` command (and any subcommands) from the `influxdb` repo.

danxmoran:
Closes #9

Most of the interesting logic was already implemented in the local config service

### DIFF
--- a/cmd/influx/main.go
+++ b/cmd/influx/main.go
@@ -35,15 +35,15 @@ func init() {
 }
 
 var (
-	tokenFlag       = "token"
-	hostFlag        = "host"
-	skipVerifyFlag  = "skip-verify"
-	traceIdFlag     = "trace-debug-id"
-	configPathFlag  = "configs-path"
-	configNameFlag  = "active-config"
-	httpDebugFlag   = "http-debug"
-	printJsonFlag   = "json"
-	hideHeadersFlag = "hide-headers"
+	tokenFlagName       = "token"
+	hostFlagName        = "host"
+	skipVerifyFlagName  = "skip-verify"
+	traceIdFlagName     = "trace-debug-id"
+	configPathFlagName  = "configs-path"
+	configNameFlagName  = "active-config"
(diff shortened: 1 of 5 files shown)


cmd/influx/main.go
```diff
@@ -35,15 +35,15 @@ func init() {
 }
 
 var (
-	tokenFlag       = "token"
```
danxmoran, line 38: Renamed these for clarity since I needed to make a top-level variable with type `cli.Flag`

internal/cmd/config/config_test.go
```diff
@@ -0,0 +1,247 @@
...
+
+	cli := config.Client{CLI: cmd.CLI{ConfigService: svc, StdIO: stdio}}
+	require.NoError(t, cli.PrintActive())
+	testutils.MatchLines(t, []string{
+		`Active\s+Name\s+URL\s+Org`,
+		fmt.Sprintf(`\*\s+%s\s+%s\s+%s`, cfg.Name, cfg.Host, cfg.Org),
+	}, strings.Split(writtenBytes.String(), "\n"))
+}
+
+func TestClient_Create(t *testing.T) {
```
lesam, line 72: These tests feel a bit too coupled to the internal implementation of the client - if we're going to have real integration tests I think we could just drop these, I doubt they would save more time catching bugs than they would cost updating them if we have to make changes.

lesam (approved):
See comment - I think the tests are not too valuable, but overall lgtm

internal/cmd/config/config_test.go
danxmoran: Will keep them for now since they're the only tests we've got, but agreed that these aren't very useful. I don't expect anything in here will change drastically before the 2.1.0 release so we should (hopefully) avoid any annoyance around changing things until after that point.

PR #94, opened 2021-05-14
refactor: replace global slices and flags with functions

danxmoran:
I started to see strange behavior where flags would merge across subcommands while working on #13.

I'm not sure if the bug was in our `append` usage or in `urfave/cli`'s handling of flag-slices, but this change seems to fix the problem either way.


cmd/influx/global.go
```diff
@@ -0,0 +1,275 @@
+package main
```
danxmoran, line 1: This file is the renamed `params.go`. It previously contained `getOrgBucketFlags`. Everything else that's added here was moved from `main.go`

wbaker85 (approved):
Interesting! The fix looks good.

PR #95, opened 2021-05-17
refactor(api): move CLI-specific API contract into this repo, add openapi as submodule

danxmoran:
This makes iterating on changes to the CLI contract _slightly_ easier, but the main motivation is to establish a slightly better separation of concerns between the repos. `openapi` is meant to be the source-of-truth for our server-side APIs, so external clients/docs can point to a single place. The fact that the CLI has a separate YML contract is more of an implementation detail of the codegen system, so tracking it alongside the server-side contracts hasn't felt "right" to me.

Bundling `openapi` as a git submodule allows us to link the common YML components from the local filesystem without maintaining a true copy of everything in `influx-cli`.

See influxdata/openapi#64 for the effects of removing the YMLs from that repo.


.gitmodules
```diff
@@ -0,0 +1,3 @@
+[submodule "openapi"]
+	path = internal/api/contract/openapi
+	url = ../openapi
```
danxmoran, line 3: Using a relative URL here is a trick that allows cloning the submodule using either `https` or `ssh`, depending on how you cloned the `influx-cli` repo.

wbaker85:
A couple high-level comments/questions which are mostly speculative ....@glinton might have thoughts on the future of the `openapi` repo regarding e2e testing in particular:

- On the subject of using a git submodule, I could imagine a scenario where you might be modifying the spec in the submodule (which is actually the `openapi` repo) alongside working on the CLI in an isolated way. You might get everything working well with the CLI and the changes in the submodule, but then when you go to push the submodule changes to the `openapi` repo, you find that these changes break something somewhere else that you hadn't anticipated and you have to circle back around. An alternative to the submodule is to make a branch directly on the `openapi` repo, make your changes there & push them to a branch, and then update the `generate-openapi.sh` script to _temporarily_ reference the commit on that branch while you work on the CLI. That's extremely clunky and not very appealing right now, but I'm thinking about a future where there might be e2e tests that run on branches in `openapi` to verify that changes to the contracts don't break any other consumers. If that were to come to pass, you would get faster feedback that the change you want to make for the contract consumed by the CLI is incompatible with other users.

- I'm also curious how other consumers of the `openapi` repo get the job done. The only one I've seen directly is the `ui` repo, which references it by commit and pulls it in that way. Are any other consumers using a submodule? My thoughts above are probably already addressed by them if there are other consumers using it in that one.

- I can see why having the CLI-specific overrides/customizations in the CLI repo makes sense...but also...that's going to make it more difficult for somebody editing the `openapi` repo directly to know if what they are doing is breaking the CLI. This is again where e2e testing that does not yet exist might come in to play: If somebody is editing a contract in `openapi`, and e2e test could show that it breaks the CLI. Then if the CLI overrides/customizations are in the `openapi` repo as well, they might be able to fix the overrides along with the contract changes. e2e testing would still show that the contract changes break the CLI if the cli-specific stuff is in the CLI repo...but then changes would need to be made to two separate repos in order to get the job done.

danxmoran:
Agreed in general that without e2e tests we're at risk of unexpected breakages in both directions.

I think there are three separate design questions to answer:
1. Should `influx-cli` pin a specific version/commit of `openapi`, or pull from `master`?
2. Should CLI-specific API customizations live in `openapi`, or in `influx-cli`?
3. Should `influx-cli` pull common pieces from `openapi` as a git submodule, or as a download over HTTP?

For question 1, I hope everyone agrees that pinning a specific version/commit is the right move. IMO this insulates the CLI a bit from breaking changes made to `openapi` (though it'd still be nice for an e2e test to catch breaking changes early).

For question 2, I like hosting customizations in `influx-cli` because it feels like a better division of responsibilities at face value, and IMO it further insulates the CLI from any breaking changes made to `openapi`. I expect that in the long run the CLI will need to make very few changes to the core `openapi` repo, and instead most of the API-related changes will be to extend the `cli.yml` file to reference additional routes and schemas. If `cli.yml` is tracked in `influx-cli` we can add new routes without needing to bump the `openapi` commit we pull from, which gives us some breathing room in the off chance that a big breaking change is ever merged to `master` there.

Question 3 is a bit linked to question 2. If everything is in `openapi`, then a submodule is overkill. On the other hand, if some components are hosted in `influx-cli` then a submodule is an easy way to handle tracking & downloading all the component pieces of `openapi`. For now I like the submodule because it makes testing `openapi` changes easier, but in the long run I don't expect that'll be a major factor.

A future-looking point I can think of in favor of the submodule is that if we ever want to write automation that can flexibly check if a given `openapi` reference (commit, tag, or branch) is compatible with a given `influx-cli` reference, `cd`-ing into the submodule and checking out the reference is a lot cleaner than a `sed` command etc. to swap the commit referenced by a download script. That said, the counterpoints could be: a submodule will add a new layer of complexity to the git-flow for the repo, and if we did write a `sed` to automate swapping a commit reference it'd probably be stable enough that we don't need to care if it's messy.

wbaker85:
Good points - no doubt on question 1, and for question 2 it's definitely helpful to think about the majority of the CLI-driven API changes being specific to the customizations rather than the underlying API definition, so it makes sense to have the customizations be in the CLI repo so the CLI can be iterated independently.

I don't have any fundamental issue with the submodule approach again considering the the CLI is probably not going to drive changes in the underlying API definition very often, and it seems like the workflow of how they will be used is sufficiently documented & automated with what you're proposing in this PR. It will be interesting to see how e2e testing turns out between different repos using `openapi` - maybe using submodules will end up being an advantage!

internal/api/contract/overrides/schemas/Extern.yml
```diff
@@ -0,0 +1,12 @@
+description: Free-form Flux AST to prepend to query requests
+type: object
+properties:
+  type:
+    type: string
+    enum: [File]
+    default: File
+# NOTE: Purposefully type-less here because the boilerplate produced
```
wbaker85, line 8:
```suggestion
# NOTE: Intentionally type-less here because the boilerplate produced
```
Alternatively, I think "purposely" would work, but for some reason the spelling of that word just looks off to me :smile:

wbaker85 (approved):
Looks good to me!

PR #96, opened 2021-05-17
feat: port top-level `org` commands from `influxdb`.

danxmoran:
Part of #13 

Not very interesting, CRUD logic + printing

### DIFF
--- a/internal/cmd/org/org.go
+++ b/internal/cmd/org/org.go
@@ -28,7 +28,7 @@ func (c Client) Create(ctx context.Context, params *CreateParams) error {
 func (c Client) Delete(ctx context.Context, id influxid.ID) error {
 	org, err := c.GetOrgsID(ctx, id.String()).Execute()
 	if err != nil {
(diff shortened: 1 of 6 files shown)


internal/cmd/org/org.go
```diff
@@ -0,0 +1,84 @@
...
+	if err != nil {
+		return fmt.Errorf("failed to create org %q: %w", params.Name, err)
+	}
+	return c.printOrgs(printOrgOpts{org: &res})
+}
+
+func (c Client) Delete(ctx context.Context, id influxid.ID) error {
+	org, err := c.GetOrgsID(ctx, id.String()).Execute()
+	if err != nil {
+		return fmt.Errorf("org %q not found", id)
```
davidby-influx, line 31: %w to wrap causal error here?

davidby-influx (changes requested):
One change to help us debug problems....

davidby-influx (approved):
LGTM

PR #98, opened 2021-05-19
feat: port `delete` from `influxdb`

### LINKED ISSUE #11: Port the `delete` command tree from `influxdb`
Port the `delete` command (and any subcommands) from the `influxdb` repo.

danxmoran:
Closes #11


internal/api/cli.gen.yml
```diff
@@ -0,0 +1,1987 @@
+openapi: 3.0.0
```
danxmoran, line 1: Oops, this one shouldn't be checked in

internal/api/contract/cli.yml
```diff
@@ -33,6 +33,8 @@ paths:
     $ref: "./overrides/paths/query.yml"
```
danxmoran, line 33: Need to add a `responses` section to handle changes made in `openapi`

internal/cmd/delete/delete.go
```diff
@@ -0,0 +1,69 @@
...
+)
+
+func (c Client) Delete(ctx context.Context, params *Params) error {
+	if !params.OrgID.Valid() && params.OrgName == "" && c.ActiveConfig.Org == "" {
+		return ErrMustSpecifyOrg
+	}
+	if !params.BucketID.Valid() && params.BucketName == "" {
+		return ErrMustSpecifyBucket
+	}
+	start, err := time.Parse(time.RFC3339Nano, params.Start)
```
wbaker85, line 37: I had a couple synapses connect while thinking through the time parsing. A while back I worked on this: https://github.com/influxdata/influxdb/pull/21369 - which added a range of allowable start and stop times that the `/api/v2/delete` API endpoint. When you get into the inner workings of the OSS storage engine the dates get converted to unix time and there can be some weird stuff happening if the dates are outside a (admittedly ridiculous) timeframe.

So I'm wondering what happens if you send dates outside this time range via the CLI to the OSS API. I guess you'd get an error response that the CLI would display? That would probably be alright. Or do you need to prevent invalid time ranges on the CLI...and how that might work with cloud, since I don't know if that API has similar limitations.

danxmoran: Good thought, looks like a decent UX to me:
```
$ ./bin/darwin/influx delete --bucket dan --start '1677-09-21T00:12:42Z' --stop '2020-01-01T00:00:00Z'
Error: failed to delete data: invalid request; error parsing request json: invalid start time, start time must not be before 1677-09-21T00:12:43.145224194Z
$ ./bin/darwin/influx delete --bucket dan --start '2020-01-01T00:00:00Z' --stop '2262-04-11T23:47:17Z'
Error: failed to delete data: invalid request; error parsing request json: invalid stop time, stop time must not be after 2262-04-11T23:47:16.854775806Z
```

wbaker85: Good deal, that does seem to convey the right message.

wbaker85 (approved):
:+1:

PR #99, opened 2021-05-20
feat: port `user` commands from `influxdb`

### LINKED ISSUE #22: Port the `user` command tree from `influxdb`
Port the `user` command (and any subcommands) from the `influxdb` repo.

danxmoran:
Closes #22


internal/cmd/user/user.go
```diff
@@ -0,0 +1,224 @@
...
+}
+
+type CreateParams struct {
+	cmd.OrgParams
+	Name     string
+	Password string
+}
+
+var (
+	ErrMustSpecifyOrg  = errors.New("must specify org ID or org name")
```
wbaker85, line 26: Looks like there's an identical error defined here as well: https://github.com/influxdata/influx-cli/blob/e4af676ae57e51b2cea7b47498d877630de8df76/internal/cmd/delete/delete.go#L26

I feel like I've seen the org and user errors used elsewhere too maybe. Would it be worth moving them somewhere that they could be defined once and shared? Kind of like with the `ErrPasswordIsTooShort`.

danxmoran: Good point, I'll do some consolidation

PR #101, opened 2021-05-21
build: set up goreleaser for release (cross-)builds

### LINKED ISSUE #100: Set up goreleaser cross-build
We should be able to build for (at least):
* `linux/amd64`
* `linux/arm64`
* `darwin/amd64`
* `windows/amd64`

1. Write the goreleaser config to build these binaries
2. Update CI to run the cross-build

danxmoran:
Closes #100 

I left out all the packaging fanciness for now.


.circleci/config.yml
```diff
@@ -43,7 +43,7 @@ jobs:
     working_directory: /home/circleci/go/src/github.com/influxdata/influx-cli
     steps:
       - checkout
-      - run: make influx
+      - run: make crossbuild
```
danxmoran, line 46: `crossbuild` was _very_ fast for me locally, checking if it's the same in CI (if not, we might want to split into parallelized jobs)

danxmoran: `16s` for this vs. `2s` for the last `make influx` to run on the `main` branch

Makefile
```diff
@@ -38,14 +38,34 @@ fmt: $(SOURCES_NO_VENDOR)
...
-	$(GO_BUILD) -o $@ ./cmd/$(shell basename "$@")
+	CGO_ENABLED=0 $(GO_BUILD) -o $@ ./cmd/$(shell basename "$@")
 
 .DEFAULT_GOAL := influx
 influx: bin/$(GOOS)/influx
 
 vendor: go.mod go.sum
 	go mod vendor
 
+GORELEASER_VERSION := v0.165.0
```
codyshepherd, line 49: Is there a process in place to keep this updated?

.circleci/config.yml
```diff
@@ -43,7 +43,9 @@ jobs:
...
+    store_artifacts:
```
codyshepherd: I believe this needs a `-` in front of it to make it one of the steps of the job

codyshepherd (approved):
I would like to see a plan for keeping the goreleaser version updated in the makefile, especially since we are still using pre-release and it seems likely to see a fair amount of churn.

But for now this is not a blocker.

Makefile
danxmoran: Not really... my thought was to keep things stable and only upgrade when we wanted new features, but if we want to always grab the latest-and-greatest when we could omit the `VERSION` tag entirely and have the download script find & get the latest available each time it runs

PR #102, opened 2021-05-21
feat: create subcommand for tasks cli

lesam:
Partial close of https://github.com/influxdata/influx-cli/issues/19

This implements the `task create` subcommand


cmd/influx/global.go
```diff
@@ -242,34 +246,94 @@ func commonFlags() []cli.Flag {
...
+// the flags to the given params container.
+func getOrgBucketFlags(c *cmd.OrgBucketParams) []cli.Flag {
+	return append(getBucketFlags(&c.BucketParams), getOrgFlags(&c.OrgParams)...)
+}
+
+
+// readQuery reads a Flux query into memory from a file, args, or stdin based on CLI parameters.
+func readQuery(ctx *cli.Context) (string, error) {
+	nargs := ctx.NArg()
+	file := ctx.String("file")
```
danxmoran, line 297: Feels like it'd be "cleaner" to have the flag name be a function arg, but if all the call-sites use the same name then maybe it's overkill? WDYT?

cmd/influx/task.go
```diff
@@ -0,0 +1,105 @@
...
+import (
+	"github.com/influxdata/influx-cli/v2/internal/cmd/task"
+	"github.com/influxdata/influx-cli/v2/pkg/cli/middleware"
+	"github.com/urfave/cli/v2"
+)
+
+func newTaskCommand() *cli.Command {
+	return &cli.Command{
+		Name:  "task",
+		Usage: "User management commands",
```
danxmoran, line 12:
```suggestion
		Usage: "Task management commands",
```

cmd/influx/task.go
```diff
@@ -0,0 +1,105 @@
...
+
+func newTaskCreateCmd() *cli.Command {
+	var params task.CreateParams
+	flags := append(commonFlags(), getOrgFlags(&params.OrgParams)...)
+	flags = append(flags, &cli.StringFlag{
+		Name:        "file",
+		Usage:       "Path to Flux script file",
+		Aliases:     []string{"f"},
+		TakesFile:   true,
+		Destination: new(string),
```
danxmoran, line 33: Should be ok to omit `Destination` if you're not trying to write into a local var

cmd/influx/task.go
```diff
@@ -0,0 +1,106 @@
...
+			params.FluxQuery, err = readQuery(ctx)
+			if err != nil {
+				return err
+			}
+			return client.Create(ctx.Context, &params)
+		},
+	}
+}
+
+/*
```
danxmoran, line 57: Reminder to delete

internal/cmd/task/task.go
```diff
@@ -0,0 +1,99 @@
...
+
+	"github.com/influxdata/influx-cli/v2/internal/api"
+	"github.com/influxdata/influx-cli/v2/internal/cmd"
+)
+
+type Client struct {
+	cmd.CLI
+	api.TasksApi
+	// AllowEmptyOrg will be useful for Kapacitor which doesn't use org / orgID
+	AllowEmptyOrg bool
```
danxmoran, line 14: Hmm, on the `Client` because it'll be useful across commands?

internal/cmd/user/user.go
```diff
@@ -24,30 +24,38 @@ type CreateParams struct {
 
 var ErrMustSpecifyUser = errors.New("must specify user ID or user name")
 
-func (c Client) Create(ctx context.Context, params *CreateParams) error {
+func getOrgID(ctx context.Context, params *cmd.OrgParams, c cmd.CLI, orgApi api.OrganizationsApi) (string, error) {
```
danxmoran, line 27: Why these changes?

cmd/influx/global.go
lesam: I initially thought this, but then realised the error messages are all specific to flux queries too, so we should probably keep the args consistent for all flux queries in the cli

internal/cmd/user/user.go
lesam: Was just reading this file to understand how to deal with orgs, and it was more obvious to read this way. I think it's likely that this function will eventually be useful elsewhere, but haven't found anywhere else to use it yet

internal/cmd/task/task.go
lesam: Each command seems to deal with org indepedently - I wouldn't want to put this on the client unless I re-evaluated all the other org fetching code, which seems like a bigger change.

danxmoran: Ah, let me rephrase: it's on `task.Client` instead of `task.CreateParams` because it'll be useful for other task functions?

danxmoran (approved):

PR #105, opened 2021-05-25
feat: task subcommand

### LINKED ISSUE #19: Port the `task` command tree from `influxdb`
Port the `task` command (and any subcommands) from the `influxdb` repo.

lesam:
Closes #19


clients/task/task.go
```diff
@@ -52,9 +54,217 @@ func (c Client) Create(ctx context.Context, params *CreateParams) error {
...
+	// If we get an id, just find the one task
+	if params.TaskID != "" {
+		task, err := c.GetTasksID(ctx, params.TaskID).Execute()
+		if err != nil {
+			return err
+		}
+		tasks = append(tasks, task)
+	} else {
+		// filter on all tasks
+		taskGet := c.GetTasks(ctx).Org(org).Limit(int32(params.Limit))
```
danxmoran, line 83: We could add `format: int64` to the schema for the limit parameter if we want the codegen to generate that. It looks like the hand-written server code uses `int` so (I think) it should be safe: https://github.com/influxdata/influxdb/blob/master/paging.go#L33-L34

If we want to keep the API as-is for now, could you move the conversion out a level and type `Limit` as an `int32` in `FindParams`?

danxmoran, line 83: Is the task API one of the "smart" ones that can handle either an ID or a name in the `org` parameter? The swagger spec doesn't mention it (but I wouldn't be shocked to learn that it's been implemented)

clients/task/task.go
```diff
@@ -52,9 +54,217 @@ func (c Client) Create(ctx context.Context, params *CreateParams) error {
...
+	return c.printTasks(taskPrintOpts{
+		tasks: tasks,
+	})
+}
+
+func (c Client) appendRuns(ctx context.Context, prev []api.Run, taskID string, filter RunFilter) ([]api.Run, error) {
+	if filter.Limit < 1 {
+		return nil, fmt.Errorf("must specify a positive run limit, not %d", filter.Limit)
+	}
+	getRuns := c.GetTasksIDRuns(ctx, taskID).Limit(int32(filter.Limit))
```
danxmoran, line 102: Same comments about `int32` and `Limit`

clients/task/task.go
```diff
@@ -52,9 +54,217 @@ func (c Client) Create(ctx context.Context, params *CreateParams) error {
...
+	}
+	var taskIds []string
+	if params.TaskID != "" {
+		taskIds = []string{params.TaskID}
+	} else {
+		org, err := c.getOrg(&params.OrgParams)
+		if err != nil {
+			return err
+		}
+		tasks, err := c.GetTasks(ctx).Limit(int32(params.TaskLimit)).Org(org).Execute()
```
danxmoran, line 181: Same comments as above:
* `int32` and `Limit`
* Will `Org(org)` work if `org` is an ID?

cmd/influx/task.go
```diff
@@ -53,3 +55,312 @@ func newTaskCreateCmd() *cli.Command {
 		},
 	}
 }
+
+func newTaskFindCmd() *cli.Command {
+	var params task.FindParams
+	params.Limit = TaskMaxPageSize
+	flags := append(commonFlags(), getOrgFlags(&params.OrgParams)...)
+	flags = append(flags, []cli.Flag{
+		&cli.StringFlag{
```
danxmoran, line 64: Non-blocking since it's already inconsistent across commands in this repo, but curious to hear your thoughts:

It looks like these are `platform.ID`s on the backend. Stuart ported that type into this repo as `influxid.ID` so we could push the validation to the "edge" of CLI parsing ([for example](https://github.com/influxdata/influx-cli/blob/main/cmd/influx/org.go#L65-L71)). I've been trying to weigh the cost-benefit of that when all the CLI does with IDs is call `.String()` on them. What do you think?

cmd/influx/task.go
```diff
@@ -53,3 +55,312 @@ func newTaskCreateCmd() *cli.Command {
 		},
 	}
 }
+
+func newTaskFindCmd() *cli.Command {
+	var params task.FindParams
+	params.Limit = TaskMaxPageSize
```
danxmoran, line 61: I think that instead of setting it here, you could set it as `Value` on the `cli.IntFlag` struct. That way it'll show up in the CLI help-text

cmd/influx/task.go
```diff
@@ -53,3 +55,312 @@ func newTaskCreateCmd() *cli.Command {
...
+			}
+			return client.Find(ctx.Context, &params)
+		},
+	}
+}
+
+func newTaskRetryFailedCmd() *cli.Command {
+	var params task.RetryFailedParams
+	params.RunFilter.Limit = 100
+	params.TaskLimit = 100
```
danxmoran, line 102: Same comment for these 2 lines, I think `Value` will have the same affect + nicer CLI help text

cmd/influx/task.go
```diff
@@ -53,3 +55,312 @@ func newTaskCreateCmd() *cli.Command {
...
+		Subcommands: []*cli.Command{
+			newTaskRunFindCmd(),
+			newTaskRunRetryCmd(),
+		},
+	}
+}
+
+func newTaskRunFindCmd() *cli.Command {
+	var params task.RunFindParams
+	params.Filter.Limit = 100
```
danxmoran, line 285: Same here

clients/task/task.go
lesam: No I don't believe it is - is that important?

cmd/influx/task.go
lesam: I'm a bit worried that if we every changed things the CLI would stop working (e.g. we change the ID length, allow a special value of some kind, etc) that would be hard to update.

I would be much happier if we were able to validate it at the openapi spec level - can we constrain and check fields on the API to ensure they are id's?

clients/task/task.go
lesam: I can't move the conversion out, and I think int64 is way overkill for these limits - for now I'll check and error if the value doesn't fit in 32 bits, before conversion.

danxmoran: Yes, I think it'll fail as-is if somebody passes `--org-id`. I think there needs to be a conditional that checks:
* If `--org-id` given, set `OrgId`
* Else set `Org` using either `--org` or the config profile

cmd/influx/task.go
danxmoran: Probably, if it's just a length limit & set of allowable characters... I can investigate

clients/task/task.go
lesam: You're right - testing showed this was broken. It should be good now - see the new commit.

danxmoran (approved):
LGTM, thanks for this!

PR #112, opened 2021-06-10
refactor: return full HTTP response for binary response bodies

danxmoran:
Implementing the multipart decoder needed for `influx backup` requires more than just the `io.ReadCloser` body from the response. This refactors the existing code in preparation of adding the new command, to keep the diff small.


wbaker85 (approved):
LGTM!

PR #116, opened 2021-06-10
feat: reimplement `influx backup` using new APIs

### LINKED ISSUE #6: Port the `backup` command tree from `influxdb`
Port the `backup` command (and any subcommands) from the `influxdb` repo.

NOTE: Use the new API design from influxdata/openapi#96

### LINKED ISSUE #113: Allow users to disable gzipping local files produced by `influx backup`
See the original community post: https://community.influxdata.com/t/disable-gzip-for-database-backups/20110

Add a `--compression` toggle to `influx backup` with options `none` and `gzip` (default `gzip`). Gzip-or-not the files written to disk based on the value of the flag.

danxmoran:
Closes #6
Closes #113 

Tests aren't comprehensive, but I think they cover the most interesting pieces of the logic.


clients/backup/backup.go
```diff
@@ -0,0 +1,283 @@
...
+// manifests from the server. KV and SQL are written to local files. Bucket manifests
+// are parsed into a slice for additional processing.
+func (c *Client) downloadMetadata(ctx context.Context, params *Params) error {
+	log.Println("INFO: Downloading metadata snapshot")
+	rawResp, err := c.GetBackupMetadata(ctx).AcceptEncoding("gzip").Execute()
+	if err != nil {
+		return fmt.Errorf("failed to download metadata snapshot: %w", err)
+	}
+
+	kvName := fmt.Sprintf("%s.bolt", c.baseName)
```
wbaker85, line 94: We are setting a base name on the server as well and included it with the response (https://github.com/influxdata/influxdb/blob/40c7d694154e75103f614cc15fdf329cdb5c0bcc/http/backup_service.go#L163). Looking at the client implementation, I think it's better to set it here though - easier to make sure all the backed up files match.

Unless you need it for something on the client (I don't think so?), at some point we might want to remove setting the filename on the server response, just to keep things from being confusing.

wbaker85 (approved):
No blockers, looks good!

clients/backup/backup.go
danxmoran: :+1: I think removing it would make sense. I opened an issue to support gzipping the response on the server-side, I'll add a note about dropping the file names there as well.

PR #117, opened 2021-06-11
test: remove racy logic from backup test

danxmoran:
Parallel tests were calling `log.SetOutput`. I considered making the tests run sequentially, but I think there are suites in other packages that also modify the global `log` state and I didn't want to just push the problem down the road. Asserting on the log output is something I feel ok not doing.


wbaker85 (approved):
👍

PR #118, opened 2021-06-11
refactor: move request-compression out of codegen, avoid some buffering

danxmoran:
`influx backup` writes files to local disk with gzip compression. `influx restore` wants to send those files back to the server, using gzip compression in transit. The existing codegen auto-applied `gzip` compression any time the `Content-Encoding: gzip` header was set; to integrate with this logic, the `restore` command would have needed to decompress the local files for the codegen to re-compress.

This PR moves compression out of the codegen entirely. Commands will need to manually gzip data when needed, but this opens up the possibility for commands to pass already-compressed data from the file system directly to the server.

I also modified the codegen to skip buffering request bodies in-memory when the body-type is easily convertible to an `io.ReadCloser`.


wbaker85 (approved):
👍

PR #119, opened 2021-06-11
refactor: use io.ReadCloser instead of **os.File for binary request params

danxmoran:
This will make POST-ing files w/ gzip slightly more convenient, since we can plug a `ReadCloser` directly into the request instead of writing a temp file.

### DIFF
--- a/api/templates/api.mustache
+++ b/api/templates/api.mustache
@@ -53,7 +53,7 @@ type {{#structPrefix}}{{&classname}}{{/structPrefix}}Api{{operationId}}Request s
 	ApiService *{{classname}}Service
 {{/generateInterfaces}}
 {{#allParams}}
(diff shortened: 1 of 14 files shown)


api/templates/api.mustache
```diff
@@ -42,21 +42,6 @@ type {{classname}} interface {
...
-}
-func (gzrc *{{#lambda.camelcase}}{{classname}}{{/lambda.camelcase}}GzipReadCloser) Read(p []byte) (int, error) {
-	return gzrc.gzip.Read(p)
-}
-func (gzrc *{{#lambda.camelcase}}{{classname}}{{/lambda.camelcase}}GzipReadCloser) Close() error {
-	if err := gzrc.gzip.Close(); err != nil {
-		return err
-	}
-	return gzrc.underlying.Close()
-}
```
danxmoran, line 58: This should have been deleted in the previous cleanup

wbaker85 (approved):

PR #120, opened 2021-06-14
feat: reimplement `influx telegrafs` with new APIs

### LINKED ISSUE #20: Port the `telegrafs` command tree from `influxdb`
Port the `telegrafs` command (and any subcommands) from the `influxdb` repo.

DStrand1:
Closes #20

There were no test cases for this originally, so I still need to do that. However, I wanted to open this PR now to get some reviews over the main code as I'm writing tests for it


clients/telegrafs/telegrafs.go
```diff
@@ -0,0 +1,173 @@
...
+
+	graf, err := c.PutTelegrafsID(ctx, id.String()).TelegrafRequest(updateTelegraf).Execute()
+	if err != nil {
+		return err
+	}
+
+	return c.writeTelegrafConfigs(telegrafPrintOpts{graf: &graf})
+}
+
+func (c Client) writeTelegrafConfigs(opts telegrafPrintOpts) error {
```
danxmoran, line 126:
```suggestion
func (c Client) printTelegrafs(opts telegrafPrintOpts) error {
```
Slightly more clear IMO. I wasn't very familiar with this command so when I saw `writeTelegrafConfigs` it might be writing the full TOML configs to local disk.

clients/telegrafs/telegrafs.go
```diff
@@ -0,0 +1,173 @@
...
+	Ids []string
+}
+
+type telegrafPrintOpts struct {
+	graf  *api.Telegraf
+	grafs *api.Telegrafs
+}
+
+func (c Client) Telegrafs(ctx context.Context, params *TelegrafParams) error {
+	if params.OrgID == 0 || params.Id == "" {
```
danxmoran, line 36: Needs a check on `params.Org` as well? (or maybe it's `params.OrgName`)

danxmoran: You can also check `!params.Valid()` instead of checking for `0`

clients/telegrafs/telegrafs.go
```diff
@@ -0,0 +1,173 @@
...
+	if file != "" {
+		bb, err := ioutil.ReadFile(file)
+		if err != nil {
+			return "", err
+		}
+
+		return string(bb), nil
+	}
+
+	bb, err := ioutil.ReadAll(os.Stdin)
```
danxmoran, line 168: Ideally we could use `c.StdIO` instead of `os.Stdin`, to help with unit testing. To do that we'd need to add `io.Reader` to the `StdIO` interface, and then plumb through the implementation.

Up to you if you want to do it as part of this PR, or create a follow-up ticket.

cmd/influx/telegrafs.go
```diff
@@ -0,0 +1,157 @@
...
+	})
+	return &cli.Command{
+		Name:  "telegrafs",
+		Usage: "List Telegraf configuration(s). Subcommands manage Telegraf configurations.",
+		Subcommands: []*cli.Command{
+			newCreateTelegrafCmd(),
+			newRemoveTelegrafCmd(),
+			newUpdateTelegrafCmd(),
+		},
+		Flags: flags,
```
danxmoran, line 28: All the `cli.Command` structs in here need a `Before: middleware.WithBeforeFns(withCli(), withApi(true))` to ensure dependencies are properly set up. Otherwise `getAPI` and `getCLI` will panic at runtime.

danxmoran (commented):
High-level structure looks good! Some initial thoughts

clients/telegrafs/telegrafs.go
```diff
@@ -0,0 +1,173 @@
...
+	Id  string
+	Ids []string
+}
+
+type telegrafPrintOpts struct {
+	graf  *api.Telegraf
+	grafs *api.Telegrafs
+}
+
+func (c Client) Telegrafs(ctx context.Context, params *TelegrafParams) error {
```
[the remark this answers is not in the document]
DStrand1: Yeah I definitely agree, I was noticing this issue more when writing test cases

clients/telegrafs/telegrafs.go
DStrand1: Looks like in the original code, it was getting the OrgID from a function by passing in both the ID and Name, and `if id == 0`, then it would lookup by name. Do we have functionality somewhere globally for this, or should I implement it myself?

clients/telegrafs/telegrafs.go
DStrand1: I think it makes sense to do this as a separate PR since it will end up touching a lot more of the code base, like the mock IO code gen

clients/telegrafs/telegrafs.go
```diff
@@ -0,0 +1,187 @@
...
+	Name string
+}
+
+func (c Client) Create(ctx context.Context, params *CreateParams) error {
+	cfg, err := c.readConfig(params.File)
+	if err != nil {
+		return err
+	}
+
+	orgID := params.OrgID.String()
```
danxmoran, line 69: The telegrafs command also supports creating within an org by name. In that case you'll need to run a GET /orgs to find the org ID. See the bucket-create code for an example: https://github.com/influxdata/influx-cli/blob/581daaa5baf33f17a2dc2cdcd0ce6e8c8561bc9e/clients/bucket/create.go#L62-L76

clients/telegrafs/telegrafs.go
```diff
@@ -0,0 +1,187 @@
...
+}
+
+type RemoveParams struct {
+	clients.OrgParams
+	Ids []string
+}
+
+func (c Client) Remove(ctx context.Context, params *RemoveParams) error {
+	for _, rawID := range params.Ids {
+		id, err := influxid.IDFromString(rawID)
```
danxmoran, line 93: I don't think there's much point in these conversions since we're just calling `String()` on them below. The server will reject any malformed IDs when the request is sent.

clients/telegrafs/telegrafs.go
```diff
@@ -0,0 +1,187 @@
...
+	Id string
+}
+
+func (c Client) List(ctx context.Context, params *ListParams) error {
+	if !params.OrgID.Valid() || params.Id == "" {
+		return errors.New("at least one of org, org-id, or id must be provided")
+	}
+
+	if params.Id != "" {
+		id, err := influxid.IDFromString(params.Id)
```
danxmoran, line 35: I don't think we need to do this conversion, since we're just calling `String()` on it below.

clients/telegrafs/telegrafs.go
```diff
@@ -0,0 +1,187 @@
...
+	Id   string
+}
+
+func (c Client) Update(ctx context.Context, params *UpdateParams) error {
+	cfg, err := c.readConfig(params.File)
+	if err != nil {
+		return err
+	}
+
+	id, err := influxid.IDFromString(params.Id)
```
danxmoran, line 119: Would remove this conversion step

clients/telegrafs/telegrafs.go
```diff
@@ -0,0 +1,187 @@
...
+	graf, err := c.PostTelegrafs(ctx).TelegrafRequest(newTelegraf).Execute()
+	if err != nil {
+		return err
+	}
+
+	return c.printTelegrafs(telegrafPrintOpts{graf: &graf})
+}
+
+type RemoveParams struct {
+	clients.OrgParams
```
danxmoran, line 87: It doesn't look like this is needed for the command to execute, could we remove it? I'll comment on the flag definitions w/ a way to preserve backwards compatibility for existing calls that use these flags.

clients/telegrafs/telegrafs.go
```diff
@@ -0,0 +1,187 @@
...
+
+		if err = c.DeleteTelegrafsID(ctx, id.String()).Execute(); err != nil {
+			return err
+		}
+	}
+	return nil
+}
+
+type UpdateParams struct {
+	clients.OrgParams
```
danxmoran, line 106: We don't typically support moving resources between orgs via an `update` command... not sure if this was added intentionally to the old CLI, or if it was an oversight from copying patterns in other commands. I'm pretty sure auth will get mismatched on the server-side if this is used.

Could we remove these params here?

clients/telegrafs/telegrafs.go
```diff
@@ -0,0 +1,187 @@
...
+			"Name":        u.GetName(),
+			"Description": u.GetDescription(),
+		}
+		rows = append(rows, row)
+	}
+
+	return c.PrintTable(headers, rows...)
+}
+
+func (c Client) readConfig(file string) (string, error) {
```
danxmoran, line 172: On 2nd look at this: Since this method is doing some either-or juggling between a CLI flag and stdin, I think it'd be better to be defined & called from `cmd/influx/telegrafs.go` instead of here. The params that currently take a `File string` would then instead take a `Config string` containing the contents of the config file. That's the pattern used by `query` and `task` to handle reading Flux queries from either file or stdin.

cmd/influx/telegrafs.go
```diff
@@ -0,0 +1,158 @@
...
+	flags := append(commonFlags(), getOrgFlags(&params.OrgParams)...)
+	flags = append(flags, &cli.StringFlag{
+		Name:        "id",
+		Usage:       "Telegraf configuration ID to retrieve",
+		Aliases:     []string{"i"},
+		Destination: &params.Id,
+	})
+	return &cli.Command{
+		Name:  "telegrafs",
+		Usage: "List Telegraf configuration(s). Subcommands manage Telegraf configurations.",
```
danxmoran, line 22: The old CLI has longer usage/example strings that I think we should carry forward. You can add `Description` fields with raw/multiline strings for that. For example, here's how `influx config` does it: https://github.com/influxdata/influx-cli/blob/9747d05ae1d3b726d2935ae8230724e58eba6442/cmd/influx/config.go#L231-L248

cmd/influx/telegrafs.go
```diff
@@ -0,0 +1,158 @@
...
+				TelegrafsApi: api.TelegrafsApi,
+			}
+			return client.Create(ctx.Context, &params)
+		},
+	}
+}
+
+func newRemoveTelegrafCmd() *cli.Command {
+	var params telegrafs.RemoveParams
+	flags := commonFlags()
```
danxmoran, line 82: Ah, looks like the unused `OrgParams` from the client aren't even bound here.

For backwards-compatibility I'd add some minimal flags to the list you `append` below:
```go
{
    Name: "org",
    Hidden: true,
},
{
    Name: "org-id",
    Hidden: true,
}
```
That'll prevent the CLI from erroring if the flags are passed, but also remove them from the help-text so new users stop passing them.

cmd/influx/telegrafs.go
```diff
@@ -0,0 +1,158 @@
...
+		Flags:   flags,
+		Before:  middleware.WithBeforeFns(withCli(), withApi(true)),
+		Action: func(ctx *cli.Context) error {
+
+			// The old CLI allowed specifying this either via repeated flags or
+			// via a single flag w/ a comma-separated value.
+			rawIds := ctx.StringSlice("id")
+			var ids []string
+			for _, p := range rawIds {
+				ids = append(ids, strings.Split(p, ",")...)
```
danxmoran, line 101: I double-checked and found that this isn't needed for backwards-compatibility, because the telegrafs command was using cobra's `StringArrayVar` instead of its `StringSliceVar` for this flag.

cmd/influx/telegrafs.go
```diff
@@ -0,0 +1,158 @@
...
+				TelegrafsApi: api.TelegrafsApi,
+			}
+			return client.Remove(ctx.Context, &params)
+		},
+	}
+}
+
+func newUpdateTelegrafCmd() *cli.Command {
+	var params telegrafs.UpdateParams
+	flags := append(commonFlags(), getOrgFlags(&params.OrgParams)...)
```
danxmoran, line 117: If you remove the `params.OrgParams` from the `UpdateParams` as suggested, this will need to change. I'd replace the `getOrgFlags` with the minimal `Hidden` flags I suggested for the `delete` command.

cmd/influx/telegrafs.go
```diff
@@ -0,0 +1,158 @@
...
+	flags = append(flags, []cli.Flag{
```
danxmoran, line 118: Since `append` is varargs, you shouldn't need the `[]cli.Flag{ /* stuff */ }...` wrapper here.

cmd/influx/telegrafs.go
```diff
@@ -0,0 +1,158 @@
...
+			}
+			return client.List(ctx.Context, &params)
+		},
+	}
+}
+
+func newCreateTelegrafCmd() *cli.Command {
+	var params telegrafs.CreateParams
+	flags := append(commonFlags(), getOrgFlags(&params.OrgParams)...)
+	flags = append(flags, []cli.Flag{
```
danxmoran, line 44: Should be able to remove the `[]cli.Flag{}...` wrapper

clients/telegrafs/telegrafs.go
```diff
@@ -0,0 +1,187 @@
...
+
+		telegraf, err := c.GetTelegrafsID(ctx, id.String()).Execute()
+		if err != nil {
+			return err
+		}
+
+		return c.printTelegrafs(telegrafPrintOpts{graf: &telegraf})
+	}
+
+	telegrafs, err := c.GetTelegrafs(ctx).OrgID(params.OrgID.String()).Execute()
```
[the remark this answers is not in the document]
DStrand1: Yeah, I missed a piece of the old code, where it would automatically lookup OrgID based on either OrgName or active config if possible

clients/telegrafs/telegrafs.go
DStrand1: Yeah, I accidentally put this parameter into here when I reorganized the structs, `remove` doesn't need it. It wasn't listed in the original command in InfluxDB either

api/contract/cli.yml
```diff
@@ -229,3 +233,9 @@ components:
       $ref: "./openapi/src/oss/schemas/BucketShardMappings.yml"
     BucketShardMapping:
       $ref: "./openapi/src/oss/schemas/BucketShardMapping.yml"
+    Telegrafs:
+      $ref: "./openapi/src/common/schemas/Telegrafs.yml"
+    TelegrafRequest:
+      $ref: "./openapi/src/common/schemas/TelegrafRequest.yml"
+    Telegraf:
+      $ref: "./openapi/src/common/schemas/Telegraf.yml"
```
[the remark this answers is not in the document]
DStrand1: Whenever I update OpenAPI to master, there is an issue saying `no such file or directory, open '/api/contract/openapi/src/oss/paths/restore_bucket-metadata_bucketID.yml`. This file seems to have been added via [ebb70c5](https://github.com/influxdata/openapi/commit/ebb70c5c53d946fdff97c492edc21ac70ad00cfd), but doesn't seem to exist on the current master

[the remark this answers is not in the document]
DStrand1: Also it seems like it does exist on master, but under the name `restore_bucket-metadata.yml`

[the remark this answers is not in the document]
danxmoran: Ah, that's my bad. #121 has the change to make `cli.yml` match up.

cmd/influx/telegrafs.go
```diff
@@ -0,0 +1,241 @@
...
+			return client.Update(ctx.Context, &params)
+		},
+	}
+}
+
+func readConfig(file string) (string, error) {
+	if file != "" {
+		bb, err := ioutil.ReadFile(file)
+		if err != nil {
+			return "", err
```
danxmoran, line 230: Let's wrap this `err` for better reporting, something like:
```suggestion
			return "", fmt.Errorf("failed to read telegraf config from %q: %w", file, err)
```

cmd/influx/telegrafs.go
```diff
@@ -0,0 +1,241 @@
...
+		}
+
+		return string(bb), nil
+	}
+
+	bb, err := ioutil.ReadAll(os.Stdin)
+	if err != nil {
+		return "", err
```
danxmoran, line 238: And this one like:
```suggestion
		return "", fmt.Errorf("failed to read telegraf config from stdin: %w", err)
```

danxmoran (commented):
A final round of suggestions to polish things up, but the core logic LGTM.

clients/telegrafs/telegrafs.go
```diff
@@ -0,0 +1,175 @@
...
+	newTelegraf := api.TelegrafRequest{
+		Name:        &params.Name,
+		Description: &params.Desc,
+		Config:      &params.Config,
+		OrgID:       &orgID,
+	}
+
+	graf, err := c.PostTelegrafs(ctx).TelegrafRequest(newTelegraf).Execute()
+	if err != nil {
+		return fmt.Errorf("failed to create telegraf config %w", err) // todo more info
```
danxmoran, line 78:
```suggestion
		return fmt.Errorf("failed to create telegraf config %q: %w", params.Name, err)
```

clients/telegrafs/telegrafs.go
```diff
@@ -0,0 +1,175 @@
...
+	}
+
+	orgID, err := c.getOrgID(ctx, params.OrgID, params.OrgName)
+	if err != nil {
+		return err
+	}
+
+	telegrafs, err := c.GetTelegrafs(ctx).OrgID(orgID).Execute()
+	if err != nil {
+		return fmt.Errorf("failed to get telegraf config with OrgID %q: %w", orgID, err)
```
danxmoran, line 50:
```suggestion
		return fmt.Errorf("failed to get telegraf configs from org %q: %w", orgID, err)
```

clients/telegrafs/telegrafs.go
```diff
@@ -0,0 +1,175 @@
...
+}
+
+type RemoveParams struct {
+	Ids []string
+}
+
+func (c Client) Remove(ctx context.Context, params *RemoveParams) error {
+	for _, rawID := range params.Ids {
+		if err := c.DeleteTelegrafsID(ctx, rawID).Execute(); err != nil {
+			return fmt.Errorf("failed to delete telegraf config with Id %q: %w", rawID, err)
```
danxmoran, line 91:
```suggestion
			return fmt.Errorf("failed to delete telegraf config with ID %q: %w", rawID, err)
```

clients/telegrafs/telegrafs.go
```diff
@@ -0,0 +1,175 @@
...
+	Desc   string
+	Config string
+	Name   string
+	Id     string
+}
+
+func (c Client) Update(ctx context.Context, params *UpdateParams) error {
+	oldTelegraf, err := c.GetTelegrafsID(ctx, params.Id).Execute()
+	if err != nil {
+		return fmt.Errorf("failed to find telegraf config with Id %q: %w", params.Id, err)
```
danxmoran, line 107:
```suggestion
		return fmt.Errorf("failed to find telegraf config with ID %q: %w", params.Id, err)
```

clients/telegrafs/telegrafs.go
```diff
@@ -0,0 +1,175 @@
...
+	updateTelegraf := api.TelegrafRequest{
+		Name:        &params.Name,
+		Description: &params.Desc,
+		Config:      &params.Config,
+		OrgID:       orgID,
+	}
+
+	graf, err := c.PutTelegrafsID(ctx, params.Id).TelegrafRequest(updateTelegraf).Execute()
+	if err != nil {
+		return fmt.Errorf("failed to update telegraf config with Id %q: %w", params.Id, err)
```
danxmoran, line 120:
```suggestion
		return fmt.Errorf("failed to update telegraf config with ID %q: %w", params.Id, err)
```

danxmoran (approved):
Final tweaks, then :shipit:

PR #121, opened 2021-06-14
feat: implement `influx restore` using new APIs

### LINKED ISSUE #15: Port the `restore` command tree from `influxdb`
Port the `restore` command (and any subcommands) from the `influxdb` repo.

NOTE: Use the new API design from influxdata/openapi#97

danxmoran:
Closes #15

I've manually tested that this works, but still need to write some unit tests for the more interesting pieces of logic.

### DIFF
--- a/clients/restore/restore.go
+++ b/clients/restore/restore.go
@@ -246,7 +246,7 @@ func (c Client) restoreOrg(ctx context.Context, name string) (string, error) {
 		return "", fmt.Errorf("failed to check existence of organization %q: %w", name, err)
 	}
 
-	if orgs.Orgs == nil || len(*orgs.Orgs) == 0 {
+	if len(orgs.GetOrgs()) == 0 {
 		// Create any missing orgs.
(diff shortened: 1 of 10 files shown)


clients/restore/restore.go
```diff
@@ -0,0 +1,292 @@
...
+	if err != nil {
+		return fmt.Errorf("failed to find backup manifests at %q: %w", path, err)
+	} else if len(manifests) == 0 {
+		return fmt.Errorf("no backup manifests found at %q", path)
+	}
+	sort.Strings(manifests)
+
+	bucketManifests := map[string]br.ManifestBucketEntry{}
+	for _, manifestFile := range manifests {
+		// Skip file if it is a directory.
```
wbaker85, line 95: Is it possible for this to happen? I guess if there was a directory called `whatever.manifest` ?

danxmoran: Yeah, I wouldn't expect it to happen but it's in the 2.0.x implementation so I'm guessing somebody got bitten by a weird filesystem layout in the past.

wbaker85 (approved):
Looks good to me!

PR #122, opened 2021-06-15
feat: reimplement `influx dashboards` with new APIs

### LINKED ISSUE #10: Port the `dashboards` command tree from `influxdb`
Port the `dashboards` command (and any subcommands) from the `influxdb` repo.

### LINKED ISSUE #108: `influx dashboards` doesn't work when `--org` is specified
__Steps to reproduce:__
1. Initialize an influxdb OSS instance locally (I use OS X), create an organization and a bucket. For my case, I did the initialization using the http://localhost:9999 UI.
2. While the OSS is running, try to access the database using `influx repl` command:
```
➜  influxdb git:(master) influx repl -o "my-org"
Error: Unable to find organization: organization not found.
```
3. Go to the web UI, find the organization Id for "my-org", try again access the database by using the organization id:
```
➜  influxdb git:(master) influx repl --org-id 047e8606a2268000
>
```
This works.

__Expected behavior:__
The `-o` flag with the organization name should work the same as the organization id.

__Actual behavior:__
Only organization id works.

__Environment info:__

* System info: Run `uname -srm` and copy the output here
  Darwin 19.0.0 x86_64
* InfluxDB version: Run `influxd version` and copy the output here
  2.0.0 (cde064b)
  **Note**: `influxd version` does not work:
```
➜  influxdb git:(master) influxd version
Error: unknown command "version" for "influxd"
Run 'influxd --help' for usage.
```

DStrand1:
Closes #10 
Closes #108


api/contract/cli.yml
```diff
@@ -241,3 +243,13 @@ components:
...
+    Dashboards:
+      $ref: "./openapi/src/common/schemas/Dashboards.yml"
+    Dashboard:
+      $ref: "./openapi/src/common/schemas/Dashboard.yml"
+    CreateDashboardRequest:
+      $ref: "./openapi/src/common/schemas/CreateDashboardRequest.yml"
+    Cells:
+      $ref: "./openapi/src/common/schemas/Cells.yml"
+    Cell:
+      $ref: "./openapi/src/common/schemas/Cell.yml"
```
danxmoran, line 255: Submodule might be out-of-sync from the rebase/merge of the telegrafs code

clients/dashboards/dashboards.go
```diff
@@ -0,0 +1,82 @@
...
+	Ids []string
+}
+
+func (c Client) List(ctx context.Context, params *Params) error {
+	orgID, err := c.getOrgID(ctx, params.OrgID, params.OrgName)
+	if err != nil {
+		return err
+	}
+	if orgID == "" && len(params.Ids) == 0 {
+		return fmt.Errorf("at least one of org, org-id, or id must be provided")
```
danxmoran, line 29: Should this check run before `getOrgID`? I think if it runs after, `getOrgID` might search for orgs using the empty string as the org name and get back every org.

danxmoran: If this gets moved to the front, the condition will need to check `params.OrgName` and `c.ActiveConfig.Org` as well (like in the telegraf commands)

clients/dashboards/dashboards.go
```diff
@@ -0,0 +1,82 @@
...
+	api.OrganizationsApi
+}
+
+type Params struct {
+	clients.OrgParams
+	Ids []string
+}
+
+func (c Client) List(ctx context.Context, params *Params) error {
+	orgID, err := c.getOrgID(ctx, params.OrgID, params.OrgName)
```
danxmoran, line 24: It looks like `GetDashboards` supports querying by org name directly. Instead of this org ID lookup, I'd make the `req`-building code below use either `OrgID` or `Org` depending on the passed parameters. For example: https://github.com/influxdata/influx-cli/blob/abe521add0c20d91499e353e57c087d1bd56c431/clients/bucket/list.go#L23-L31

The difference can be significant based on the permissions assigned to the user running the command.

clients/dashboards/dashboards.go
```diff
@@ -0,0 +1,82 @@
...
+	req = req.OrgID(orgID).Id(params.Ids)
+	dashboards, err := req.Execute()
+	if err != nil {
+		return fmt.Errorf("failed to find dashboards with OrgID %q and IDs %q: %w", orgID, params.Ids, err)
+	}
+
+	return c.printDashboards(dashboards)
+}
+
+func (c Client) getOrgID(ctx context.Context, orgID influxid.ID, orgName string) (string, error) {
```
danxmoran, line 44: Can probably get rid of this method

clients/dashboards/dashboards.go
DStrand1: This was how it was structured before, but now that you've brought it up, I think this was part of the issue in #108, this and the fact that it didn't check orgName consistently

clients/dashboards/dashboards.go
```diff
@@ -0,0 +1,78 @@
...
+	api.OrganizationsApi
+}
+
+type Params struct {
+	clients.OrgParams
+	Ids []string
+}
+
+func (c Client) List(ctx context.Context, params *Params) error {
+	if !params.OrgID.Valid() && params.OrgName == "" && len(params.Ids) == 0 {
```
danxmoran, line 23:
```suggestion
	if !params.OrgID.Valid() && params.OrgName == "" && c.ActiveConfig.Org == "" && len(params.Ids) == 0 {
```

clients/dashboards/dashboards.go
```diff
@@ -0,0 +1,78 @@
...
+		return fmt.Errorf("at least one of org, org-id, or id must be provided")
+	}
+
+	var req api.ApiGetOrgsRequest
```
danxmoran, line 27: Ah, miscommunication. This logic is good but it can/should be running directly against the `GetDashboards` request below. I don't think there's any need to use the `GetOrgs` API in this method.

clients/dashboards/dashboards.go
```diff
@@ -0,0 +1,78 @@
...
+	"fmt"
+
+	"github.com/influxdata/influx-cli/v2/api"
+	"github.com/influxdata/influx-cli/v2/clients"
+)
+
+type Client struct {
+	clients.CLI
+	api.DashboardsApi
+	api.OrganizationsApi
```
danxmoran, line 14: Should be able to drop this embedded client

clients/dashboards/dashboards.go
```diff
@@ -0,0 +1,66 @@
...
+	if params.OrgID.Valid() {
+		req = req.OrgID(params.OrgID.String())
+	}
+	if params.OrgName != "" {
+		req = req.Org(params.OrgName)
+	}
+	if !params.OrgID.Valid() && params.OrgName == "" {
+		req = req.Org(c.ActiveConfig.Org)
+	}
+	orgID := req.GetOrgID() // for logging purposes
```
danxmoran, line 38: This will be an empty string if `--org-id` wasn't passed. I think you can drop it and simplify the error message slightly, see suggestion below

clients/dashboards/dashboards.go
```diff
@@ -0,0 +1,66 @@
...
+	dashboards, err := req.Id(params.Ids).Execute()
+	if err != nil {
+		return fmt.Errorf("failed to find dashboards with OrgID %q and IDs %q: %w", orgID, params.Ids, err)
```
danxmoran, line 41:
```suggestion
		return fmt.Errorf("failed to find dashboards: %w", err)
```

danxmoran (approved):
A final suggestion, then LGTM

clients/dashboards/dashboards.go
danxmoran: I think this will also fix the build failure

PR #123, opened 2021-06-16
refactor: move things around to make building a client easier from other projects

danxmoran:
* Move the `config` package out of `internal/` so other projects can reference it when building a `clients.CLI`
* Split the logic for building an `api.APIClient` out of `cmd/` so other projects can reuse it


api/templates/client.mustache
```diff
@@ -60,9 +60,9 @@ type service struct {
 	client *APIClient
 }
 
-// NewAPIClient creates a new API client. Requires a userAgent string describing your application.
+// newAPIClient creates a new API client. Requires a userAgent string describing your application.
 // optionally a custom http.Client to allow for advanced features such as caching.
-func NewAPIClient(cfg *Configuration) *APIClient {
+func newAPIClient(cfg *Configuration) *APIClient {
```
danxmoran, line 65: @lesam will this change break how Kapacitor is using the CLI? I can submit a PR to fix it after this merges

lesam: Yes it will break: https://github.com/influxdata/kapacitor/blob/master/cmd/kapacitor/fluxtask.go#L95

danxmoran: Reverted this, made the new helper function return `api.Configuration` instead

lesam (approved):
lgtm

api/configuration_default.go
```diff
@@ -0,0 +1,44 @@
...
+	Token            *string
+	TraceId          *string
+	AllowInsecureTLS bool
+	Debug            bool
+}
+
+// NewAPIConfig builds a configuration tailored to the InfluxDB v2 API.
+func NewAPIConfig(params ConfigParams) *Configuration {
+	clientTransport := http.DefaultTransport.(*http.Transport)
+	clientTransport.TLSClientConfig = &tls.Config{InsecureSkipVerify: params.AllowInsecureTLS}
```
lesam, line 22: Since this is likely the only use in the process it's probably fine, but should we be mutating the global DefaultTransport?

danxmoran: Updated to `Clone()` the transport first.

Issue #114, opened 2021-06-10
Consistently use `influxid.ID` (or delete it)


danxmoran:
#53 introduced the `influxid.ID` type into this project to model resource IDs. A few of our CLI commands use it, but others use plain `string`. Our codegen accepts & returns `string` IDs.

We need to decide if the extra up-front validation provided by `influxid.ID` is worth the extra ceremony needed to convert that type to/from strings in our core logic. If it's worth it, we should update all of our commands to use it. If it's not worth it, we should remove the type and convert the existing use-sites to `string`s.

PR #125, opened 2021-06-16
feat: reimplement `influx secret` CLI command

### LINKED ISSUE #16: Port the `secret` command tree from `influxdb`
Port the `secret` command (and any subcommands) from the `influxdb` repo.

DStrand1:
Closes #16

Existing test cases from InfluxDB relating to the business logic (not the externally handled things like `commonFlags`) were ported over in full.


clients/secret/input.go
```diff
@@ -0,0 +1,43 @@
+package secret
```
danxmoran, line 1: (I think) you shouldn't need to add this file. Instead, you can use the `GetPassword` method on `StdIO` that's already in the repo. If your client embeds a `clients.CLI` then you should be able to do (roughly) `c.StdIO.GetPassword`.

We might want to rename the method to `GetSecret` to show it's more generic...

go.mod
```diff
@@ -13,6 +13,7 @@ require (
 	github.com/google/go-cmp v0.5.5
 	github.com/kr/pretty v0.1.0 // indirect
 	github.com/stretchr/testify v1.7.0
+	github.com/tcnksm/go-input v0.0.0-20180404061846-548a7d7a8ee8
```
danxmoran, line 16: We (hopefully) shouldn't need to add this dependency

clients/secret/secret.go
```diff
@@ -0,0 +1,165 @@
...
+	Key string
+}
+
+func (c Client) Delete(ctx context.Context, params *DeleteParams) error {
+	orgID, err := c.getOrgID(ctx, params.OrgParams)
+	if err != nil {
+		return err
+	}
+
+	if err := c.PostOrgsIDSecrets(ctx, orgID).Execute(); err != nil {
```
danxmoran, line 38: That's an unfortunate name... can you add a comment above this saying that `PostOrgsIDSecrets` actually deletes the secret, and the name is generated from the `operationId` in our API docs?

danxmoran, line 38: I think this is missing a request body before the `Execute()`

clients/secret/secret.go
```diff
@@ -0,0 +1,165 @@
...
+		return fmt.Errorf("failed to delete secret with key %q: %v", params.Key, err)
```
danxmoran, line 39:
```suggestion
		return fmt.Errorf("failed to delete secret with key %q: %w", params.Key, err)
```

clients/secret/secret.go
```diff
@@ -0,0 +1,165 @@
...
+	}
+
+	secrets := response.GetSecrets()
+	opts := make([]secret, 0, len(secrets))
+	for _, entry := range secrets {
+		opts = append(opts, secret{
+			key:   entry,
+			orgID: orgID,
+		})
+	}
```
danxmoran, line 74: Since you know the exact length, you could instead:
```suggestion
	opts := make([]secret, len(secrets))
	for i, entry := range secrets {
		opts[i] = secret{
			key:   entry,
			orgID: orgID,
		}
	}
```

clients/secret/secret.go
```diff
@@ -0,0 +1,165 @@
...
+func (c Client) List(ctx context.Context, params *ListParams) error {
+
+	orgID, err := c.getOrgID(ctx, params.OrgParams)
+	if err != nil {
+		return err
+	}
+
+	response, err := c.GetOrgsIDSecrets(ctx, orgID).Execute()
+	if err != nil {
+		return fmt.Errorf("failed to retrieve secret keys: %s", err)
```
danxmoran, line 64:
```suggestion
		return fmt.Errorf("failed to retrieve secret keys: %w", err)
```

clients/secret/secret.go
```diff
@@ -0,0 +1,165 @@
...
+		if err != nil {
+			return err
+		}
+	}
+
+	err = c.PatchOrgsIDSecrets(ctx, orgID).RequestBody(map[string]string{
+		params.Key: secretVal,
+	}).Execute()
+	if err != nil {
+		return fmt.Errorf("failed to update secret with key %q: %v", params.Key, err)
```
danxmoran, line 104:
```suggestion
		return fmt.Errorf("failed to update secret with key %q: %w", params.Key, err)
```

clients/secret/secret.go
```diff
@@ -0,0 +1,165 @@
...
+	} else {
+		secretVal, err = c.StdIO.GetSecret("Please type your secret", 0)
+		if err != nil {
+			return err
+		}
+	}
+
+	err = c.PatchOrgsIDSecrets(ctx, orgID).RequestBody(map[string]string{
+		params.Key: secretVal,
+	}).Execute()
```
danxmoran, line 102: Formatting nitpick:
```suggestion
	err = c.PatchOrgsIDSecrets(ctx, orgID).
		RequestBody(map[string]string{params.Key: secretVal}).
		Execute()
```

clients/secret/secret.go
```diff
@@ -0,0 +1,165 @@
...
+			v = opts.secret
+		}
+		return c.PrintJSON(v)
+	}
+
+	headers := []string{"Key", "Organization ID"}
+	if opts.deleted {
+		headers = append(headers, "Deleted")
+	}
+	if opts.secrets == nil {
```
danxmoran, line 128: It probably doesn't matter since we only use 1 or the other, but I think the pattern used in other commands is instead:
```suggestion
	if opts.secret != nil {
```

DStrand1: I'm not totally understanding why, but when I attempt to make this change, my IDE is telling me that `nil cannot be reassigned to type "secret"`. This almost feels like an IDE error, though there is probably a good explanation

danxmoran: Ah, it's because you've typed `opts.secret` as `secret` instead of `*secret`

clients/secret/secret_test.go
```diff
@@ -0,0 +1,94 @@
...
+		{
+			name: "org id",
+			params: secret.ListParams{
+				OrgParams: clients.OrgParams{
+					OrgID: id,
+				},
+			},
+			defaultOrgName: "default-org",
+			registerExpectations: func(t *testing.T, secretApi *mock.MockSecretsApi) {
+				secretApi.EXPECT().GetOrgsIDSecrets(gomock.Any(), gomock.Any()).Return(api.ApiGetOrgsIDSecretsRequest{ApiService: secretApi})
```
danxmoran, line 42: Instead of `gomock.Any` here for the 2nd arg, you could use `gomock.Eq(id.String())`

clients/secret/secret_test.go
```diff
@@ -0,0 +1,94 @@
...
+	"github.com/stretchr/testify/assert"
+	tmock "github.com/stretchr/testify/mock"
+	"github.com/stretchr/testify/require"
+	"testing"
+)
+
+func TestSecret_List(t *testing.T) {
+	t.Parallel()
+
+	id, err := influxid.IDFromString("1111111111111111")
```
danxmoran, line 22: Since everything uses this as a `string`, you could skip the `IDFromString` wrapper logic

clients/secret/secret_test.go
```diff
@@ -0,0 +1,94 @@
...
+				},
+			},
+			defaultOrgName: "default-org",
+			registerExpectations: func(t *testing.T, secretApi *mock.MockSecretsApi) {
+				secretApi.EXPECT().GetOrgsIDSecrets(gomock.Any(), gomock.Any()).Return(api.ApiGetOrgsIDSecretsRequest{ApiService: secretApi})
+				secretApi.EXPECT().GetOrgsIDSecretsExecute(tmock.MatchedBy(func(in api.ApiGetOrgsIDSecretsRequest) bool {
+					body := in.GetOrgID()
+					return assert.NotEmpty(t, body) &&
+						assert.Equal(t, body, id.String())
+				})).Return(api.SecretKeysResponse{Secrets: &[]string{fakeResults}}, nil)
```
danxmoran, line 47: Since we expect `Execute` to take the request generated by `GetOrgsIDSecrets` w/ no changes, this can be simplified:
```suggestion
				req := api.ApiGetOrgsIDSecretsRequest{ApiService: secretApi}.OrgID(id)
				secretApi.EXPECT().GetOrgsIDSecrets(gomock.Any(), gomock.Eq(id)).Return(req)
				secretApi.EXPECT().GetOrgsIDSecretsExecute(gomock.Eq(req)).Return(api.SecretKeysResponse{Secrets: &[]string{fakeResults}}, nil)
```

clients/secret/secret_test.go
```diff
@@ -0,0 +1,94 @@
...
+			if tc.registerExpectations != nil {
+				tc.registerExpectations(t, secretsApi)
+			}
+			cli := secret.Client{
+				CLI: clients.CLI{
+					ActiveConfig: config.Config{
+						Org: tc.defaultOrgName,
+					},
+					StdIO: stdio,
+				},
```
danxmoran, line 86: Needs to set `SecretsApi: secretsApi` to avoid the panic

clients/secret/secret_test.go
DStrand1: The `clients.OrgParams` struct still requires a type `influxid.ID` instead of String. So I think this is related to #114, but for now I have to stick with the influxid.ID type

DStrand1:
Resolved merge conflict, and added one more test case for testing default org

danxmoran (approved):
:shipit:

PR #131, opened 2021-06-17
feat: port top-level `export` command from `influxdb`

danxmoran:
Addresses part of #12 (still need to add the subcommands).

Most of the strange logic here is in the CLI parsing. I've replicated all of the existing behavior except for one case where filters were silently being converted/ignored; that case is now an error.


api/contract/cli.yml
```diff
@@ -253,3 +255,11 @@ components:
       $ref: "./openapi/src/common/schemas/Cells.yml"
     Cell:
       $ref: "./openapi/src/common/schemas/Cell.yml"
+    TemplateExport:
+      $ref: "./overrides/schemas/TemplateExport.yml"
+    Template:
+      $ref: "./openapi/src/common/schemas/Template.yml"
```
codyshepherd, line 261: Is it a problem that these `openapi/src/` paths aren't present in `api/contract`?

cmd/influx/export.go
```diff
@@ -0,0 +1,379 @@
...
+	influx export --stack-id $STACK_ID --buckets $BUCKET_ID
+
+All of the resources are supported via the examples provided above. Provide the
+resource flag and then provide the IDs.
+
+For information about exporting InfluxDB templates, see
+https://docs.influxdata.com/influxdb/latest/reference/cli/influx/export/`,
+		Flags: append(
+			commonFlagsNoPrint(),
+			&cli.StringFlag{
```
codyshepherd, line 69: minor thing, but these could be sorted in lex order to make them easier to find

codyshepherd (approved):

cmd/influx/export.go
danxmoran: Opened #133 to address this in general (it's a problem across commands).

api/contract/cli.yml
danxmoran: Hmmm where did you see the paths weren't present? `openapi` is a git submodule so it takes a `git submodule update --init` for the paths to appear locally. The UI renders the folder as a link to the other repo.

PR #134, opened 2021-06-21
feat: port `export all` command from `influxdb`

danxmoran:
Part 2/3 of #12 

Most of the diff is moving around code from the top-level `export` command.

### DIFF
--- a/clients/export/export.go
+++ b/clients/export/export.go
@@ -11,6 +11,7 @@ import (
 type Client struct {
 	clients.CLI
 	api.TemplatesApi
+	api.OrganizationsApi
 }
 
 type Params struct {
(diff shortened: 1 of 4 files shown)


clients/export/export.go
```diff
@@ -128,27 +118,86 @@ func (c Client) Export(ctx context.Context, params *Params) error {
...
+	KindFilters  []ResourceType
+}
+
+func (c Client) ExportAll(ctx context.Context, params *AllParams) error {
+	if params.OrgId == "" && params.OrgName == "" && c.ActiveConfig.Org == "" {
+		return clients.ErrMustSpecifyOrg
+	}
+
+	orgId := params.OrgId
+	if orgId == "" {
```
wbaker85, line 143: Does the server require an org ID specifically and not work with an org name?

danxmoran: For this API, yes

wbaker85 (approved):
LGTM!

PR #136, opened 2021-06-21
feat: v1 dbrp commands 

### LINKED ISSUE #24: Port the `v1 dbrp` command tree from `influxdb`
Port the `v1 dbrp` command (and any subcommands) from the `influxdb` repo.

wbaker85:
Closes #24

This adds the `v1 dbrp` subcommands for doing CRUD actions on a DBRP.

I manually tested these against both an OSS instance and on cloud. The unit tests don't cover getting an org name from the config object, but it worked fine in manual testing.


clients/v1_dbrps/v1_dbrps.go
```diff
@@ -0,0 +1,246 @@
...
+type dbrpPrintOpts struct {
+	dbrp  *api.DBRP
+	dbrps []api.DBRP
+}
+
+type ListParams struct {
+	clients.OrgParams
+	ID       string
+	BucketID string
+	Default_ bool
```
danxmoran, line 26: Why the trailing underscore here? And in the other structs

clients/v1_dbrps/v1_dbrps.go
```diff
@@ -0,0 +1,246 @@
...
+	}
+	if params.Default_ {
+		reqBody.Default = &params.Default_
+	}
+
+	req := c.PatchDBRPID(ctx, params.ID)
+	if params.OrgID.Valid() {
+		req = req.OrgID(params.OrgID.String())
+	}
+	if params.OrgName != "" {
```
danxmoran, line 150: Does the cloud API support update-by-org-name? Or do we need to look up the org ID here as well?

clients/v1_dbrps/v1_dbrps.go
```diff
@@ -0,0 +1,246 @@
...
+		return clients.ErrMustSpecifyOrg
+	}
+
+	// Get the DBRP to verify that it exists, and to be able to print the results
+	// of the delete command, which output the details of the deleted DBRP.
+	getReq := c.GetDBRPsID(ctx, params.ID)
+	deleteReq := c.DeleteDBRPID(ctx, params.ID)
+	if params.OrgID.Valid() {
+		getReq = getReq.OrgID(params.OrgID.String())
+		deleteReq = deleteReq.OrgID(params.OrgID.String())
```
danxmoran, line 184: I'm surprised either of these requests need org information to work. Can we double-check that the `org` and `orgId` params are actually used on the server-side in OSS and Cloud?

A common bug I've seen in the old CLI is that somebody copy-pasted the org/org-id flags into their command but never actually used the values. If that's the case here, we can update our CLI-parsing layer to keep the flags but mark them as hidden.

cmd/influx/v1_dbrps.go
```diff
@@ -0,0 +1,180 @@
+package main
```
danxmoran, line 1: Nitpick: I'd have kept the file name as `v1_dbrp` to match the command name 1:1

wbaker85: Changed it back to `v1_dbrp` to match the command name! I was aiming for consistency but there's a mix of singular and plural no matter what so matching the command name seems best!

clients/v1_dbrps/v1_dbrps.go
wbaker85: Partially a hold-over from when I was using lower-cased property names (which didn't work, of course) and a conflict with the reserved `default` keyboard, and partially because the codegen apparently names the property `Default_`. I changed all the property names to just be `Default` and added a comment to where the property is assigned to the codegen'd request object.

clients/v1_dbrps/v1_dbrps.go
wbaker85: Updating with an org name does work for this. It's strange that creating with an org name doesn't work for cloud - I think it's probably a bug. I opened an issue in IDPE for that.

clients/v1_dbrps/v1_dbrps.go
wbaker85: You do need the `org` or `orgID` parameter for OSS `GET/{id}` and `DELETE/{id}` requests. I believe it's because of how the authorization is set up. I suspect it wouldn't _need_ to be set up like that - I'd think the authz could work based on the org information of the requested DBRP, since that object will have an organization ID and then it should be able to auth the request against that. Might require a bit more investigation as to why it works like it does...I opened an issue in`influxdb` regarding this: https://github.com/influxdata/influxdb/issues/21731

Interestingly, requests to cloud don't require the org info for `GET/{id}` or `DELETE/{id}` I don't know if this is actually intentional since the code looks to be the same. I think it's probably how the org information gets set on the request context in IDPE for every request even if it's not provided as a query parameter.

danxmoran (approved):

PR #137, opened 2021-06-21
refactor: remove resource-type enums from `export` commands

danxmoran:
My logic for this change is:
* The CLI doesn't run different logic based on resource-type, so it only "needs" to parse the type into an enum to validate that it's in the allowed list
* The server also validates the resource-types sent to it by the client
* Removing validation from the CLI side of things loosens the coupling between CLI and server
  * This could help us avoid work in the future as new resource-types are added to the server, since we won't necessarily need to re-codegen just to pick up new enum values


davidby-influx (approved):
LGTM

PR #138, opened 2021-06-22
feat: port `influx v1 auth` command

### LINKED ISSUE #23: Port the `v1 auth` command tree from `influxdb`
Port the `v1 auth` command (and any subcommands) from the `influxdb` repo.

DStrand1:
Closes #23

There is still a few issues with this, primarily in `List` where the token field is not available. I think a lot of this is using the v2 API endpoints, but I need help looking through that and updating OpenAPI to do that. However, I believe the logic is good so updating the API calls in the logic should be trivial


api/contract/cli.yml
```diff
@@ -81,6 +81,10 @@ paths:
     $ref: "./openapi/src/common/paths/dbrps.yml"
   "/dbrps/{dbrpID}":
     $ref: "./openapi/src/common/paths/dbrps_dbrpID.yml"
+  /authorizations:
```
danxmoran, line 84: This will combine with the default server URL to become `/api/v2/authorizations` (and similar for below). The V1 auth system uses `/private/legacy/authorizations` as its endpoint prefix.

We can fix it by declaring custom `servers` for these 2 routes (see `/health` at the top for an example). I'd suggest doing it like this:
```
/legacy/authorizations:
  servers:
    - /private
```
We could also try having `/private/legacy` be the server and `/authorizations` be the operation, but I think that'll collide with V2 auth when we add it.

clients/params.go
```diff
@@ -18,3 +20,24 @@ type OrgBucketParams struct {
 	OrgParams
 	BucketParams
 }
+
+type AuthLookupParams struct {
+	ID       influxid.ID
+	Username string
+	Required bool // Required when set to true determines whether validate expects either ID or Username to be set
```
danxmoran, line 27: Instead of putting this in the struct, could we remove it and not call `Validate()` in contexts where the args aren't required?

clients/v1_auth/v1_auth.go
```diff
@@ -0,0 +1,428 @@
...
+
+	// verify an existing token with the same username doesn't already exist
+	auth, err := c.AuthorizationsApi.GetAuthorizationsID(ctx, ids.GetUsers()[0].GetId()).Execute()
+	authPtr := &auth
+	if &authPtr != nil {
+		return fmt.Errorf("authorization with username %q already exists", params.Username)
+	}
+	if err != nil {
+		return fmt.Errorf("failed to verify username %q has no auth: %w", params.Username, err)
+	}
```
danxmoran, line 68: I think these two checks are incompatible / need reworked. The query from line 55 is correct, but we want to see a 404 / empty array response to prove we aren't about to hit a name collision. The `GetAuthorizationsID` call based on ID shouldn't be needed.

You can use logic [like this](https://github.com/influxdata/influx-cli/blob/main/clients/backup/backup.go#L211-L212) to check if a non-`nil` `err` is a 404 response.

clients/v1_auth/v1_auth.go
```diff
@@ -0,0 +1,428 @@
...
+
+	password := params.Password
+	if password == "" && !params.NoPassword {
+		password, err = c.CLI.StdIO.GetSecret("Please enter your password", 0)
```
danxmoran, line 72: For a 1:1 match with the old CLI, we should:
1. Enforce a min length on the password here
2. Ask for a confirmation of the password afterwards, and ensure the values are the same

Something [like this](https://github.com/influxdata/influx-cli/blob/main/clients/setup/setup.go#L167-L184).

clients/v1_auth/v1_auth.go
```diff
@@ -0,0 +1,428 @@
...
+			// verify the input ID
+			if _, err := influxid.IDFromString(p); err != nil {
+				return fmt.Errorf("invalid bucket ID '%s': %w (did you pass a bucket name instead of an ID?)", p, err)
+			}
+
+			newPerm := api.Permission{
+				Action: bp.action,
+				Resource: api.PermissionResource{
+					Type:  "buckets",
+					Id:    *api.NewNullableString(&p),
```
danxmoran, line 98: The `*api.NewNullableString(...)` wrapper feels unnecessary here... can we use `p` and `orgID` directly?

clients/v1_auth/v1_auth.go
```diff
@@ -0,0 +1,428 @@
...
+	}
+
+	authReq := &api.Authorization{
+		Description: &params.Desc,
+		OrgID:       orgID,
+		Permissions: permissions,
+		Token:       &params.Username,
+	}
+
+	// TODO Something wrong here
```
danxmoran, line 113: What error are you seeing?

clients/v1_auth/v1_auth.go
```diff
@@ -0,0 +1,428 @@
...
+	newAuth, err := c.AuthorizationsApi.PostAuthorizations(ctx).AuthorizationPostRequest(authReq).Execute()
+	if err != nil {
+		return fmt.Errorf("failed to create new authorization: %w", err)
+	}
+
+	if password != "" {
+		if err := c.UsersApi.PostUsersIDPassword(ctx, newAuth.GetUserID()).
```
danxmoran, line 120: We don't want to use the `UsersApi` here (it manages v2 users).

Based on [this code](https://github.com/influxdata/influxdb/blob/master/v1/authorization/http_server.go#L636-L664) from the old CLI, it looks like we need to add another route to our `cli.yml` for setting v1 passwords.

clients/v1_auth/v1_auth.go
```diff
@@ -0,0 +1,428 @@
...
+			PasswordResetBody(api.PasswordResetBody{
+				Password: password,
```
danxmoran, line 122: IMO would read better if you put the whole `PasswordResetBody` + the closing `).` on one line

clients/v1_auth/v1_auth.go
DStrand1: This `AuthorizationRequestBody` type doesn't have a token field but we need it according to the old design

api/contract/cli.yml
```diff
@@ -81,6 +81,14 @@ paths:
...
+    servers:
+      - url: "/legacy"
+    $ref: "./openapi/src/common/paths/authorizations.yml"
+  /authorizations/{authID}:
+    servers:
+      - url: "/legacy"
+    $ref: "./openapi/src/common/paths/authorizations_authID.yml"
```
danxmoran, line 91: This still won't resolve to the right thing, it'll become `/legacy/authorizations/<stuff>`. This should do it:
```suggestion
  /legacy/authorizations:
    servers:
      - url: "/private"
    $ref: "./openapi/src/common/paths/authorizations.yml"
  /legacy/authorizations/{authID}:
    servers:
      - url: "/private"
    $ref: "./openapi/src/common/paths/authorizations_authID.yml"
```

clients/v1_auth/v1_auth.go
```diff
@@ -0,0 +1,428 @@
...
+}
+
+func (c Client) Remove(ctx context.Context, params *RemoveParams) error {
+	id, err := c.getAuthReqID(ctx, &params.AuthLookupParams)
+	if err != nil {
+		return err
+	}
+
+	auth, err1 := c.AuthorizationsApi.GetAuthorizationsID(ctx, id).Execute()
+	err2 := c.AuthorizationsApi.DeleteAuthorizationsID(ctx, id).Execute()
```
danxmoran, line 163: We shouldn't send this if `err1 != nil`

clients/v1_auth/v1_auth.go
```diff
@@ -0,0 +1,428 @@
...
+	if err1 != nil || err2 != nil {
+		return fmt.Errorf("could not find Authorization from ID %q: %w", id, err1)
```
danxmoran, line 165: This is the error to show if `err1 != nil`, we should have a different error if `err2 != nil`

clients/v1_auth/v1_auth.go
```diff
@@ -0,0 +1,428 @@
...
+	}
+	if params.OrgID.Valid() {
+		req = req.OrgID(params.OrgID.String())
+	}
+	if params.OrgName != "" {
+		req = req.Org(params.OrgName)
+	}
+	var id string
+	if params.AuthLookupParams.IsSet() {
+		newId, err := c.getAuthReqID(ctx, &params.AuthLookupParams)
```
danxmoran, line 213: (I think) this will make an API call based on the ID and fail if nothing is found, which isn't what we want here. Instead, we want to set ID on the filter if it's set, and name/token on the filter if it's set. The server will "handle" nonsense IDs by returning no results.

clients/v1_auth/v1_auth.go
DStrand1: We would have to update the yaml's to allow for it since it requires this type currently

clients/v1_auth/v1_auth.go
```diff
@@ -0,0 +1,428 @@
...
+		if err != nil {
+			return fmt.Errorf("could not get User from Username %q: %w", params.Username, err)
+		}
+		id = newId
+	}
+	if params.UserID != "" {
+		id = params.UserID
```
danxmoran, line 220: I don't think this is correct. `params.UserID` is the ID of the v2 user that owns the token, so it should get set directly on the request as `req.UserID`.

clients/v1_auth/v1_auth.go
```diff
@@ -0,0 +1,428 @@
...
+	}
+	if id != "" {
+		req = req.UserID(id)
```
danxmoran, line 223: This code will look very different after the filter is adjusted based on comments above, but: `id` is a v1 auth ID here, which isn't the same as `UserID`. I'd expect that:
* `req.ID` is what you should use to filter by v1 auth ID
* `req.Token` is what you should use to filter by v1 auth username

clients/v1_auth/v1_auth.go
```diff
@@ -0,0 +1,428 @@
...
+	} else {
+		status = "inactive"
+	}
+	req.AuthorizationUpdateRequest(api.AuthorizationUpdateRequest{
+		Status: &status,
+	})
+
+	auth, err := req.Execute()
+	if err != nil {
+		return fmt.Errorf("could not find Authorization with ID %q: %w", id, err)
```
danxmoran, line 280: Since this is a PATCH, not a GET:
```suggestion
		return fmt.Errorf("failed to update status of authorization with ID %q: %w", id, err)
```

clients/v1_auth/v1_auth.go
```diff
@@ -0,0 +1,428 @@
...
+	}
+
+	auth, err := c.AuthorizationsApi.GetAuthorizationsID(ctx, id).Execute()
+	if err != nil {
+		return fmt.Errorf("could not find authorization with User ID %q: %w", id, err)
+	}
+
+	password := params.Password
+	if password == "" {
+		password, err = c.StdIO.GetSecret("Please enter new password", 0)
```
danxmoran, line 324: Same comment as [here](https://github.com/influxdata/influx-cli/pull/138/files#r656473691).

clients/v1_auth/v1_auth.go
```diff
@@ -0,0 +1,428 @@
...
+		if err != nil {
+			return err
+		}
+	}
+
+	if err := c.UsersApi.PostUsersIDPassword(ctx, auth.GetUserID()).Execute(); err != nil {
```
danxmoran, line 330: Same comment as [here](https://github.com/influxdata/influx-cli/pull/138/files#r656477646).

clients/v1_auth/v1_auth.go
```diff
@@ -0,0 +1,428 @@
...
+		} else {
+			v = params.tokens
+		}
+		return c.PrintJSON(v)
+	}
+
+	headers := []string{
+		"ID",
+		"Description",
+		"Token",
```
danxmoran, line 351:
```suggestion
		"Name / Token",
```
For 1:1 with old CLI

clients/v1_auth/v1_auth.go
```diff
@@ -0,0 +1,428 @@
...
+	if params.token != nil {
+		params.tokens = append(params.tokens, *params.token)
+	}
+
+	var rows []map[string]interface{}
+	for _, u := range params.tokens {
+		row := map[string]interface{}{
+			"ID":          u.ID,
+			"Description": u.Description,
+			"Token":       u.Token,
```
danxmoran, line 368:
```suggestion
			"Name / Token":       u.Token,
```

clients/v1_auth/v1_auth.go
```diff
@@ -0,0 +1,428 @@
...
+	ret := p.GetAction() + ":"
+	r := p.GetResource()
+
+	if r.GetOrgID() != "" {
+		ret += "orgs" + r.GetOrgID()
+	}
+	ret += r.GetType()
+	if r.GetId() != "" {
+		ret += r.GetId()
+	}
```
danxmoran, line 391: There needs to be a `/` between each of the components in this for a 1:1 match

clients/v1_auth/v1_auth.go
```diff
@@ -0,0 +1,428 @@
...
+	return ret
+}
+
+func (c Client) getAuthReqID(ctx context.Context, params *clients.AuthLookupParams) (id string, err error) {
+	if params.ID.Valid() {
+		id = params.ID.String()
+	} else {
+		userid, err := c.UsersApi.GetUsers(ctx).Name(params.Username).Execute()
```
danxmoran, line 399: We should be using the `AuthorizationApi` here, not the `UsersApi`

clients/v1_auth/v1_auth.go
```diff
@@ -0,0 +1,428 @@
...
+		if err != nil || len(userid.GetUsers()) == 0 {
+			err = fmt.Errorf("could not find User ID from username %q: %w", params.Username, err)
```
danxmoran, line 401:
```suggestion
			err = fmt.Errorf("could not find v1 auth with username %q: %w", params.Username, err)
```

clients/v1_auth/v1_auth.go
```diff
@@ -0,0 +1,428 @@
...
+		} else {
+			id = userid.GetUsers()[0].GetId()
+		}
+	}
+	return
+}
+
+func (c Client) getOrgID(ctx context.Context, params clients.OrgParams) (string, error) {
+	if params.OrgID.Valid() || params.OrgName != "" || c.ActiveConfig.Org != "" {
```
danxmoran, line 410: (Sorry if I haven't caught it in past PRs)

More conventional Go would be to:
1. Invert the conditional here
2. Return the error as the body
3. Leave the remaining logic un-indented after the `if`

[For example](https://github.com/influxdata/influx-cli/blob/main/clients/bucket/create.go#L24-L28)

DStrand1:
At this point, I think we have to start diverging more heavily from the v2 auth API in our legacy API for 1.x, as these last few requested changes are unable to be done with the current API

DStrand1:
Latest push requires matching OpenAPI PR https://github.com/influxdata/openapi/pull/150

clients/params.go
```diff
@@ -18,3 +20,21 @@ type OrgBucketParams struct {
 	OrgParams
 	BucketParams
 }
+
+type AuthLookupParams struct {
```
danxmoran, line 24: I'd only expect this to be used across the v1_auth subcommands, so I'd suggest keeping it in `cmd/influx/v1_auth.go`

cmd/influx/global.go
```diff
@@ -276,3 +276,20 @@ func getBucketFlags(params *clients.BucketParams) []cli.Flag {
 func getOrgBucketFlags(c *clients.OrgBucketParams) []cli.Flag {
 	return append(getBucketFlags(&c.BucketParams), getOrgFlags(&c.OrgParams)...)
 }
+
+// getAuthLookupFlags returns flags used for authorization, requiring either an ID or Username,
+// and can specify to require one but not both.
+func getAuthLookupFlags(params *clients.AuthLookupParams) []cli.Flag {
```
danxmoran, line 282: I'd also move this to `cmd/influx/v1_auth.go` since it's not useful to other commands

clients/v1_auth/v1_auth.go
```diff
@@ -0,0 +1,423 @@
...
+			err = fmt.Errorf("could not find v1 auth with token (username) %q: %w", params.Username, err)
+		} else {
+			id = auths.GetAuthorizations()[0].GetId()
+		}
+	}
+	return
+}
+
+func (c Client) getOrgID(ctx context.Context, params clients.OrgParams) (string, error) {
+	if params.OrgID.Valid() || params.OrgName != "" || c.ActiveConfig.Org != "" {
```
danxmoran, line 406:
```suggestion
	if !params.OrgID.Valid() && params.OrgName == "" && c.ActiveConfig.Org == "" {
```
The current logic causes the a failure if org is specified through any one of the 3

clients/v1_auth/v1_auth.go
```diff
@@ -0,0 +1,423 @@
...
+	// verify an existing token with the same username doesn't already exist
+	auths, err := c.LegacyAuthorizationsApi.GetAuthorizations(ctx).Token(params.Username).Execute()
+	if apiError, ok := err.(api.ApiError); ok {
+		if apiError.ErrorCode() != api.ERRORCODE_NOT_FOUND {
+			return fmt.Errorf("failed to verify username %q has no auth: %w", params.Username, err)
+		} else if len(auths.GetAuthorizations()) != 0 {
+			return fmt.Errorf("authorization with username %q already exists", params.Username)
+		}
+	}
+	auth := auths.GetAuthorizations()[0]
```
danxmoran, line 64: This line will always fail, since we're asserting above that `len(auths.GetAuthorizations()) == 0`

clients/v1_auth/v1_auth.go
```diff
@@ -0,0 +1,423 @@
...
+				},
+			}
+			permissions = append(permissions, newPerm)
+		}
+	}
+	authReq := api.AuthorizationPostRequest{
+		Description: &params.Desc,
+		OrgID:       orgID,
+		Permissions: permissions,
+		UserID:      auth.UserID,
```
danxmoran, line 106: This will need fixed up since we don't have `auth` (because of comment above).

It looks like the old CLI doesn't even set `UserID` (maybe it's derived from the auth token on the request?). You'll need to set `Token: params.Username` here

clients/v1_auth/v1_auth.go
```diff
@@ -0,0 +1,441 @@
...
+		return fmt.Errorf("failed to create new authorization: %w", err)
+	}
+
+	if password != "" {
+		err := c.LegacyAuthorizationsApi.SetAuthorizationsIDPassword(ctx, newAuth.GetId()).
+			Password(password).
+			Execute()
+		if err != nil {
+			_ = c.LegacyAuthorizationsApi.DeleteAuthorizationsID(ctx, newAuth.GetId()).Execute()
+			return fmt.Errorf("failed to update password for user ID %q: %w", newAuth.GetUserID(), err)
```
danxmoran, line 137: I think this error message would be more understandable for users:
```suggestion
			return fmt.Errorf("failed to set password for %q: %w", params.Username, err)
```

clients/v1_auth/v1_auth.go
```diff
@@ -0,0 +1,441 @@
...
+		Token:       &params.Username,
+	}
+
+	newAuth, err := c.LegacyAuthorizationsApi.PostAuthorizations(ctx).LegacyAuthorizationPostRequest(authReq).Execute()
+	if err != nil {
+		return fmt.Errorf("failed to create new authorization: %w", err)
+	}
+
+	if password != "" {
+		err := c.LegacyAuthorizationsApi.SetAuthorizationsIDPassword(ctx, newAuth.GetId()).
```
danxmoran, line 132: `api_legacy_authorizations.gen.go` needs to be updated to use `POST` instead of `PATCH`

danxmoran: I also see it putting the password in the query params of the request. The server expects it to be in the request body, in a JSON object with 1 field of "password": https://github.com/influxdata/influxdb/blob/master/v1/authorization/http_server.go#L632-L634

clients/v1_auth/v1_auth.go
```diff
@@ -0,0 +1,443 @@
...
+	}
+
+	req := c.LegacyAuthorizationsApi.PatchAuthorizationsID(ctx, id)
+	var status string
+	if active {
+		status = "active"
+	} else {
+		status = "inactive"
+	}
+	req.AuthorizationUpdateRequest(api.AuthorizationUpdateRequest{
```
danxmoran, line 284: This needs to be:
```suggestion
	req = req.AuthorizationUpdateRequest(api.AuthorizationUpdateRequest{
```

clients/v1_auth/v1_auth.go
```diff
@@ -0,0 +1,443 @@
...
+	if params.Username != "" {
+		req = req.Token(params.Username)
+	}
+	if params.ID.Valid() {
+		req = req.AuthID(params.ID.String())
+	}
+
+	auths, err := req.Execute()
+	if err != nil {
+		return fmt.Errorf("could not find Authorizations for specified arguments %w", err)
```
danxmoran, line 238:
```suggestion
		return fmt.Errorf("could not find Authorizations for specified arguments: %w", err)
```

clients/v1_auth/v1_auth.go
```diff
@@ -0,0 +1,443 @@
...
+func (c Client) SetPassword(ctx context.Context, params *SetPasswordParams) error {
+	id, err := c.getAuthReqID(ctx, &params.AuthLookupParams)
+	if err != nil {
+		return err
+	}
+
+	auth, err := c.LegacyAuthorizationsApi.GetAuthorizationsID(ctx, id).Execute()
+	if err != nil {
+		return fmt.Errorf("could not find authorization with User ID %q: %w", id, err)
+	}
```
danxmoran, line 330: I think we can get rid of this block. It looks like the only place `auth` is used below is to get the ID, and we get that from `getAuthReqID` above.

danxmoran (approved):
🎉

PR #140, opened 2021-06-23
feat: improved error messages for cloud- or oss-only commands

### LINKED ISSUE #129: CLI: Improve error messaging when an unexpected response comes back
Sometimes, there are apis that work in OSS but not in cloud. the CLI seems to do a poor job in detecting response codes and providing a good error message. This issue is to improve that parsing and error messaging.

Here's an example:
```
[techops@t410 influxdb2-client-2.0.4-linux-amd64]$ ./influx v1 auth list --host https://us-west-2-1.aws.cloud2.influxdata.com --org-id xxxxxxxxxx --token (all access token)
Error: Invalid character '<' looking for beginning of value.
See 'influx v1 auth list -h' for help
```

wbaker85:
Closes #129

This PR adds the ability to set properties on individual API requests or all requests in a particular API service by chaining methods called `OnlyOSS()` or `OnlyCloud()` when the requests or services are created. I also tweaked the generic "undefined response type" output to be a little more descriptive.

If the request fails, the resulting error will be changed from something like this:

```
influx ping
Error: undefined response type
```

...to something like this:

```
influx ping
```

As part of this, I set `BackupApi` and `RestoreApi` to be "OnlyOSS", and the `GetHealth` request of the `Ping` command to be "OnlyOSS" as well. We could set the entire `HealthApi` to be "OnlyOSS" - I did this for demonstration purposes more than anything...and there is actually a working `/ping` endpoint on cloud so maybe we'll want to use that someday.

For cloud-only endpoints, the `BucketsSchemaApi` was set to "OnlyCloud", which changes the error message when trying to use the command on an OSS server:

```
influx bucket-schema list -n bucket
Error: failed to list measurement schemas: undefined response type
```

...becomes:

```
influx bucket-schema list -n bucket
Error: failed to list measurement schemas: InfluxDB Cloud-only command failed: unable to decode response content type "text/plain; charset=utf-8"
```

Future work, which is currently blocked by https://github.com/influxdata/influxdb/issues/20224 (& the equivalent cloud issue), could explicitly check the `X-Influxdb-Version` based on these properties and return a more specific error message. This could even include a pre-flight check of some kind to make sure the server supports the command if they are flagged as being cloud- or oss-only before trying to execute the requests. I'll create an additional issue to document this further work when we merge this PR.

### DIFF
--- a/api/templates/api.mustache
+++ b/api/templates/api.mustache
@@ -9,6 +9,7 @@ import (
 	_ioutil "io/ioutil"
 	_nethttp "net/http"
 	_neturl "net/url"
+	_fmt "fmt"
 {{#imports}}	"{{import}}"
 {{/imports}}
 )
@@ -39,11 +40,28 @@ type {{classname}} interface {
 	 */
 	{{nickname}}Execute(r {{#structPrefix}}{{&classname}}{{/structPrefix}}Api{{operationId}}Request) ({{#returnType}}{{#isResponseBinary}}*_nethttp.Response{{/isResponseBinary}}{{^isResponseBinary}}{{{returnType}}}{{/isResponseBinary}}, {{/returnType}}error)
 	{{/operation}}
+
+	// Sets the intention of the API to only work for InfluxDB OSS servers - for logging error messages
+	OnlyOSS() {{classname}}
+
+	// Sets the intention of the API to only work for InfluxDB Cloud servers - for logging error messages
+	OnlyCloud() {{classname}}
 }
 {{/generateInterfaces}}
 
 // {{classname}}Service {{classname}} service
 type {{classname}}Service service
+
+func (a *{{classname}}Service) OnlyOSS() {{classname}} {
+	a.isOnlyOSS = true
+	return a
+}
+
+func (a *{{classname}}Service) OnlyCloud() {{classname}} {
+	a.isOnlyCloud = true
+	return a
+}
+
 {{#operation}}
 
 type {{#structPrefix}}{{&classname}}{{/structPrefix}}Api{{operationId}}Request struct {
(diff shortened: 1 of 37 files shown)


api/templates/api.mustache
```diff
@@ -71,6 +95,22 @@ func (r {{#structPrefix}}{{&classname}}{{/structPrefix}}Api{{operationId}}Reques
...
+	return r
+}
+
+// Sets additional descriptive text in the error message if this specific
+// request fails, indicating that it is intended to be used only on cloud
+// servers.
+func (r {{#structPrefix}}{{&classname}}{{/structPrefix}}Api{{operationId}}Request) OnlyCloud() {{#structPrefix}}{{&classname}}{{/structPrefix}}Api{{operationId}}Request {
+	r.isOnlyCloud = true
+	return r
+}
```
danxmoran, line 112: I'd vote to keep things simple for now and have the OSS-vs-Cloud toggle at the API level, and not per-request

api/templates/api.mustache
```diff
@@ -334,20 +374,27 @@ func (a *{{{classname}}}Service) {{nickname}}Execute(r {{#structPrefix}}{{&class
...
+	} else if r.isOnlyCloud || a.isOnlyCloud {
+		errorPrefix = "InfluxDB Cloud-only command failed: "
+	}
+
 	if localVarHTTPResponse.StatusCode >= 300 {
 		body, err := GunzipIfNeeded(localVarHTTPResponse)
 		if err != nil {
 			body.Close()
-			return {{#returnType}}localVarReturnValue, {{/returnType}}err
+			return {{#returnType}}localVarReturnValue, {{/returnType}}_fmt.Errorf("%s%v", errorPrefix, err)
```
danxmoran, line 388: For these errors, I'd use `%w` instead of `%v`

api/templates/api.mustache
wbaker85: Done! Should make it easier to keep track of what has been flagged for a specific server type for sure.

api/templates/api.mustache
wbaker85: Makes sense, I updated the `fmt.Errorf` calls to use `%w` instead of `%v`- although I did have to read up on what the difference between those is!

api/templates/api.mustache
```diff
@@ -359,10 +388,10 @@ func (a *{{{classname}}}Service) {{nickname}}Execute(r {{#structPrefix}}{{&class
 			var v {{{dataType}}}
 			err = a.client.decode(&v, localVarBody, localVarHTTPResponse.Header.Get("Content-Type"))
 			if err != nil {
-				newErr.error = err.Error()
+				newErr.error = _fmt.Sprintf("%s%v", errorPrefix, err.Error())
 				return {{#returnType}}localVarReturnValue, {{/returnType}}newErr
 			}
-			newErr.model = &v
```
danxmoran, line 365: I think we need to leave this line as well, some commands rely on access to `model` to conditionally switch on error code.

wbaker85: Cool, good to know that is needed, I put it back in and everything still seems to work the same way for these error messages.

danxmoran (approved):

PR #144, opened 2021-06-23
feat: add API & override specs for template application

danxmoran:
Part of #4

I chose to override almost the entire API spec for this route.

The [`TemplateApply`](https://github.com/influxdata/openapi/blob/master/src/common/schemas/TemplateApply.yml) and [`TemplateSummary`](https://github.com/influxdata/openapi/blob/master/src/common/schemas/TemplateSummary.yml) specs tracked in `openapi` use many layers of inline objects, which is unfriendly to our codegen. Splitting layers out into separate files helps things.

Even after splitting, some schemas (i.e. [`CheckDiscriminator`](https://github.com/influxdata/openapi/blob/master/src/common/schemas/CheckDiscriminator.yml)) are complex enough to confuse the generator. I looked at those schemas and saw that the vast majority of fields aren't used by the `apply` command. The override files capture the subset of fields that are used, and collapse away any `oneOf`s.


danxmoran:
Self-TODO: Review fields and mark required things as such

wbaker85 (approved):
LGTM! (That's a lot of overrides!)

PR #145, opened 2021-06-23
feat: port table-printers from influxdb

danxmoran:
Part of #4

The diff-printer is a near-direct copy from [this code](https://github.com/influxdata/influxdb/blob/master/cmd/influx/template.go#L1787-L1939) in `influxdb`. The table-printer's [core logic](https://github.com/influxdata/influxdb/blob/master/cmd/influx/template.go#L1941-L1994) is also from `influxdb`, but I refactored it to have the same structure as the diff-printer.

### DIFF
--- a/pkg/template/diff_printer.go
+++ b/pkg/template/diff_printer.go
@@ -1,0 +1,180 @@
+package template
+
+import (
+	"fmt"
+	"io"
+	"strconv"
+	"strings"
+
+	"github.com/olekukonko/tablewriter"
+)
+
+type DiffPrinter struct {
+	w      io.Writer
+	writer *tablewriter.Table
+
+	useColor bool
+	title    string
+
+	appendCalls int
+	headerLen   int
+}
+
+func NewDiffPrinter(w io.Writer, hasColor, hasBorder bool) *DiffPrinter {
+	wr := tablewriter.NewWriter(w)
+	wr.SetBorder(hasBorder)
+	wr.SetRowLine(hasBorder)
+
+	return &DiffPrinter{
+		w:        w,
+		writer:   wr,
+		useColor: hasColor,
+	}
+}
+
+func (d *DiffPrinter) Render() {
+	if d.appendCalls == 0 {
+		return
+	}
+
+	// set the title and the add/remove legend
+	title := strings.ToUpper(d.title)
+	add := "+add"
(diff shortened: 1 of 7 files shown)


pkg/template/diff_printer.go
```diff
@@ -0,0 +1,180 @@
...
+	remove := "-remove"
+	if d.useColor {
+		title = colorTitle.Sprint(title)
+		add = colorTotalAdd.Sprint(add)
+		remove = colorTotalRemove.Sprint(remove)
+	}
+	fmt.Fprintf(d.w, "%s    %s | %s | unchanged\n", title, add, remove)
+
+	d.setFooter()
+	d.writer.Render()
```
wbaker85, line 52: There was a `fmt.Fprintln(d.w)` here in the `influxdb` code...was that not needed here?  ... or there?

danxmoran: I'd originally ported it assuming that `Render()` wasn't writing a trailing newline, but found from testing that it wasn't needed. Maybe in the `apply` output it's nice to have blank lines between each output; in that case, I'll add the println's back there.

pkg/template/diff_printer.go
```diff
@@ -0,0 +1,180 @@
...
+	"fmt"
+	"io"
+	"strconv"
+	"strings"
+
+	"github.com/olekukonko/tablewriter"
+)
+
+type DiffPrinter struct {
+	w      io.Writer
```
danxmoran, line 13: For better separation of concerns, I'm considering:
1. Making this a `bytes.Buffer`
2. Removing the `io.Writer` arg to `NewDiffPrinter`
3. Having `Render()` return a string

What do you think?

wbaker85: I do associate the term "Render" with actually printing/outputting something to the screen, so having it return a string would be a little less intuitive to me at first glance. I can see what you mean about separation of concerns though...right now `Render()` prepares the output and prints it (or writes it to `w`) both. I'm used to that though, since that's on-par with how, for example, `render()` in a React component works, more or less.

danxmoran: Hmmm ok, in that case I'll leave it as-is :+1:

wbaker85 (approved):

PR #146, opened 2021-06-23
feat: port template-parsing logic from influxdb

danxmoran:
Part of #4

`influx apply` supports reading templates from files, urls, and stdin. This is my attempt to hide these differences behind some sort of common interface.

This isn't a 1:1 match to the code from `influxdb` (I tried to clean it up some, and added tests). The original code is:
* [Here](https://github.com/influxdata/influxdb/blob/master/cmd/influx/template.go#L1027-L1066)
* [Here](https://github.com/influxdata/influxdb/blob/master/pkger/parser.go#L45-L266) (minus the logic around "source" content-type, because the derivation didn't actually work as far as I could tell)

### DIFF
--- a/clients/apply/source.go
+++ b/clients/apply/source.go
@@ -1,0 +1,215 @@
+package apply
+
+import (
+	"context"
+	"encoding/json"
+	"fmt"
+	"io"
+	"net/http"
+	"net/url"
+	"os"
+	"path"
+	"path/filepath"
+	"strings"
+
+	"github.com/influxdata/influx-cli/v2/api"
+	"github.com/influxdata/influx-cli/v2/pkg/github"
+	"github.com/influxdata/influx-cli/v2/pkg/jsonnet"
+	"gopkg.in/yaml.v3"
+)
+
+type TemplateEncoding int
+
+const (
+	TemplateEncodingUnknown TemplateEncoding = iota
+	TemplateEncodingJson
+	TemplateEncodingJsonnet
+	TemplateEncodingYaml
+)
+
+func (e *TemplateEncoding) Set(v string) error {
+	switch v {
+	case "jsonnet":
+		*e = TemplateEncodingJsonnet
+	case "json":
+		*e = TemplateEncodingJson
+	case "yml", "yaml":
+		*e = TemplateEncodingYaml
+	default:
+		return fmt.Errorf("unknown inEncoding %q", v)
+	}
+	return nil
+}
+
+func (e TemplateEncoding) String() string {
+	switch e {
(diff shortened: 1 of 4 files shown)


clients/apply/source_test.go
```diff
@@ -0,0 +1,434 @@
...
+
+	for _, tc := range testCases {
+		tc := tc
+		t.Run(tc.name, func(t *testing.T) {
+			t.Parallel()
+
+			server := httptest.NewServer(http.HandlerFunc(func(rw http.ResponseWriter, r *http.Request) {
+				rw.WriteHeader(tc.resStatus)
+				rw.Write([]byte(tc.resBody))
+				return
```
wbaker85: Looks like the linter is complaining about this; you shouldn't need a return here at the end of a http handler.

wbaker85: Ah now it's fixed, nice!

clients/apply/source.go
```diff
@@ -0,0 +1,215 @@
...
+	}
+
+	sources := make([]TemplateSource, len(paths))
+	for i := range paths {
+		path := paths[i] // Local var for the `Open` closure to capture.
+		encoding := encoding
+		if encoding == TemplateEncodingUnknown {
+			ext := filepath.Ext(path)
+			switch {
+			case strings.HasPrefix(ext, ".jsonnet"):
```
wbaker85, line 78: Any reason this switch can't be something more like:

```
switch ext {
  case ".jsonnet"
    etc...
}
```

...just not sure why it has to be a `HasPrefix` check. Maybe there's some reason for that though!

clients/apply/source.go
```diff
@@ -0,0 +1,215 @@
...
+			paths = append(paths, fullPath)
+		}
+	}
+	return paths, nil
+}
+
+func SourceFromURL(u *url.URL, encoding TemplateEncoding) TemplateSource {
+	if encoding == TemplateEncodingUnknown {
+		ext := path.Ext(u.Path)
+		switch {
```
wbaker85, line 132: I think I see why the switch statement above using the `HasPrefix` - it uses it here too for urls, which I imagine might have query parameters or other stuff (?) after the file name in the url. Makes sense to have them be the same if that's the reasoning.

wbaker85 (approved):
Looks good - one minor question, but nothing blocking!

clients/apply/source.go
danxmoran: I wouldn't expect any query parameters to show up in the post-parsing `u.Path`. I'll refactor to drop the prefix-checks

PR #148, opened 2021-06-24
fix: add yaml tags to models so we can unmarshal camelCase

danxmoran:
Was driving myself crazy trying to figure out why `apiVersion` was getting dropped from YAML templates while working on #4...

### DIFF
--- a/api/templates/model_simple.mustache
+++ b/api/templates/model_simple.mustache
@@ -1,0 +1,391 @@
+// {{classname}}{{#description}} {{{description}}}{{/description}}{{^description}} struct for {{{classname}}}{{/description}}
+type {{classname}} struct {
+{{#parent}}
+    {{^isMap}}
+        {{^isArray}}
+            {{{parent}}}
+        {{/isArray}}
+    {{/isMap}}
+    {{#isArray}}
+        Items {{{parent}}}
+    {{/isArray}}
+{{/parent}}
+{{#vars}}
+    {{^-first}}
+    {{/-first}}
+    {{#description}}
+        // {{{description}}}
+    {{/description}}
+    {{name}} {{^required}}{{^isNullable}}*{{/isNullable}}{{/required}}{{{dataType}}} `json:"{{baseName}}{{^required}},omitempty{{/required}}" yaml:"{{baseName}}{{^required}},omitempty{{/required}}"{{#withXml}} xml:"{{baseName}}{{#isXmlAttribute}},attr{{/isXmlAttribute}}"{{/withXml}}{{#vendorExtensions.x-go-custom-tag}} {{{.}}}{{/vendorExtensions.x-go-custom-tag}}`
+{{/vars}}
+{{#isAdditionalPropertiesTrue}}
+    AdditionalProperties map[string]interface{}
+{{/isAdditionalPropertiesTrue}}
+}
+
+{{#isAdditionalPropertiesTrue}}
+    type _{{{classname}}} {{{classname}}}
+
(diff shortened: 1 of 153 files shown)


api/templates/model_simple.mustache
```diff
@@ -0,0 +1,391 @@
+// {{classname}}{{#description}} {{{description}}}{{/description}}{{^description}} struct for {{{classname}}}{{/description}}
```
danxmoran, line 1: This file is copy-pasted from the source code of the generator we use. The only edit I made was to add `yaml:` tags to the templates that generate struct fields.

api/templates/model_simple.mustache
```diff
@@ -0,0 +1,391 @@
...
+        Items {{{parent}}}
+    {{/isArray}}
+{{/parent}}
+{{#vars}}
+    {{^-first}}
+    {{/-first}}
+    {{#description}}
+        // {{{description}}}
+    {{/description}}
+    {{name}} {{^required}}{{^isNullable}}*{{/isNullable}}{{/required}}{{{dataType}}} `json:"{{baseName}}{{^required}},omitempty{{/required}}" yaml:"{{baseName}}{{^required}},omitempty{{/required}}"{{#withXml}} xml:"{{baseName}}{{#isXmlAttribute}},attr{{/isXmlAttribute}}"{{/withXml}}{{#vendorExtensions.x-go-custom-tag}} {{{.}}}{{/vendorExtensions.x-go-custom-tag}}`
```
danxmoran, line 19: Here

api/templates/model_simple.mustache
```diff
@@ -0,0 +1,391 @@
...
+    {{#parent}}
+        {{^isMap}}
+            type {{classname}}WithoutEmbeddedStruct struct {
+            {{#vars}}
+                {{^-first}}
+                {{/-first}}
+                {{#description}}
+                    // {{{description}}}
+                {{/description}}
+                {{name}} {{^required}}{{^isNullable}}*{{/isNullable}}{{/required}}{{{dataType}}} `json:"{{baseName}}{{^required}},omitempty{{/required}}" yaml:"{{baseName}}{{^required}},omitempty{{/required}}"{{#withXml}} xml:"{{baseName}}{{#isXmlAttribute}},attr{{/isXmlAttribute}}"{{/withXml}}{{#vendorExtensions.x-go-custom-tag}} {{{.}}}{{/vendorExtensions.x-go-custom-tag}}`
```
danxmoran, line 289: And here

clients/apply/source_test.go
```diff
@@ -313,15 +313,15 @@ func TestTemplateSource_Read(t *testing.T) {
 	t.Parallel()
 
 	yamlTemplate := `---
-apiversion: influxdata.com/v2alpha1
+apiVersion: influxdata.com/v2alpha1
```
danxmoran, line 316: I extracted this test data by using the `export` command to marshal YAML. Since the struct tags were missing, it marshalled using the wrong field name.

wbaker85 (approved):
Looks good to me...only one though: Would you consider this to be a bug in the generator itself? Just wondering if there's already an issue for it / if it's something that should be fixed upstream.

danxmoran:
I did a quick search in their issues and couldn't find anything. It does seems like a feature others might want / they might accept, since they already have a "withXml" toggle. Leaving this open for now so I remember to open the upstream issue.

danxmoran:
OpenAPITools/openapi-generator#9855

PR #150, opened 2021-06-24
feat: port influx export stack subcommand

### LINKED ISSUE #12: Port the `export` command tree from `influxdb`
Port the `export` command (and any subcommands) from the `influxdb` repo.

mcfarlm3:
Closes #12 

Ports over the `export stack` subcommand from the influxdb repo, with modified logic to work with openapi/codegen. Usage of this command is `export stack [flags] <stack_id>`. Note that it will not currently work with the usage `export stack <stack_id> [flags]`.

### DIFF
--- a/clients/errors.go
+++ b/clients/errors.go
@@ -6,4 +6,5 @@ var (
 	ErrPasswordIsTooShort = errors.New("password is too short")
 	ErrMustSpecifyOrg     = errors.New("must specify org ID or org name")
 	ErrMustSpecifyBucket  = errors.New("must specify bucket ID or bucket name")
+	ErrMustSpecifyStack   = errors.New("must specify stack ID")
 )
--- a/clients/errors.go
+++ b/clients/errors.go
@@ -6,5 +6,4 @@ var (
 	ErrPasswordIsTooShort = errors.New("password is too short")
 	ErrMustSpecifyOrg     = errors.New("must specify org ID or org name")
 	ErrMustSpecifyBucket  = errors.New("must specify bucket ID or bucket name")
-	ErrMustSpecifyStack   = errors.New("must specify stack ID")
 )
--- a/clients/export/export.go
+++ b/clients/export/export.go
@@ -110,3 +110,9 @@ func (c Client) ExportAll(ctx context.Context, params *AllParams) error {
 	}
 	return nil
 }
+//
+//func (c Client) ExportStack(ctx context.Context, params *AllParams) error {
+//	if ctx.NArg() != 1 {
+//		return clients.ErrMustSpecifyStack
+//	}
+//}
--- a/clients/export/export.go
+++ b/clients/export/export.go
@@ -118,15 +118,14 @@ type StackParams struct {
 
 func (c Client) ExportStack(ctx context.Context, params *StackParams) error {
 	if params.StackId == "" {
-		return clients.ErrMustSpecifyStack
+		return fmt.Errorf("no stack id provided")
 	}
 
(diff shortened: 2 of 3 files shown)


clients/errors.go
```diff
@@ -6,4 +6,5 @@ var (
 	ErrPasswordIsTooShort = errors.New("password is too short")
 	ErrMustSpecifyOrg     = errors.New("must specify org ID or org name")
 	ErrMustSpecifyBucket  = errors.New("must specify bucket ID or bucket name")
+	ErrMustSpecifyStack   = errors.New("must specify stack ID")
```
danxmoran, line 9: Since this is only used by `export`, I'd put it in `export.go` for now. The other errors here cut across multiple commands.

clients/export/export.go
```diff
@@ -110,3 +110,27 @@ func (c Client) ExportAll(ctx context.Context, params *AllParams) error {
...
+	OutParams
+	StackId 	string
+}
+
+func (c Client) ExportStack(ctx context.Context, params *StackParams) error {
+	if params.StackId == "" {
+		return clients.ErrMustSpecifyStack
+	}
+
+	stackId := params.StackId
```
danxmoran, line 124: Nitpick: this local var isn't really needed, you could use `params.StackId` everywhere.

clients/export/export.go
```diff
@@ -110,3 +110,27 @@ func (c Client) ExportAll(ctx context.Context, params *AllParams) error {
...
+
+	exportReq := api.TemplateExport{StackID: &stackId}
+	tmpl, err := c.ExportTemplate(ctx).TemplateExport(exportReq).Execute()
+	if err != nil {
+		return fmt.Errorf("failed to export template: %w", err)
```
danxmoran, line 129: Would include the ID in the message:
```suggestion
		return fmt.Errorf("failed to export stack %q: %w", err)
```

clients/export/export.go
```diff
@@ -110,3 +110,27 @@ func (c Client) ExportAll(ctx context.Context, params *AllParams) error {
...
+	}
+	if err := params.OutParams.writeTemplate(tmpl); err != nil {
+		return fmt.Errorf("failed to write exported template: %w", err)
+	}
+
+	return nil
+}
```
danxmoran, line 136: Can you add a newline to the end of the file?

cmd/influx/export.go
```diff
@@ -360,6 +361,66 @@ https://docs.influxdata.com/influxdb/latest/reference/cli/influx/export/all/
...
+	var params struct {
+		out     string
+		stackId string
+	}
+
+	return &cli.Command{
+		Name:  "stack",
+		Usage: "Export all resources associated with a stack as a template",
+		Description: `The influx export stack command exports all resources 
+	associated with a stack as a template. All metadata.name fields remain the same.
```
danxmoran, line 374:
```suggestion
associated with a stack as a template. All metadata.name fields remain the same.
```

cmd/influx/export.go
```diff
@@ -360,6 +361,66 @@ https://docs.influxdata.com/influxdb/latest/reference/cli/influx/export/all/
 	}
 }
 
+func  newExportStackCmd() *cli.Command {
+	var params struct {
+		out     string
+		stackId string
```
danxmoran, line 367: (I think) you can drop this `stackId` field, it looks unused

danxmoran (commented):
Some suggestions for cleanup, but the core logic LGTM!

danxmoran (approved):

PR #151, opened 2021-06-25
fix: Accept bucket ID or bucket name arguments

### LINKED ISSUE #149: bucket-schema: Unable to list buckets using bucket-id or i option
## What

Unable to list bucket schema using the `--bucket-id` or `-i` CLI options

## Steps to reproduce issue

Execute the following command:

```sh
./bin/darwin/influx bucket-schema list -i 4d5dbdfd5a67ae4f --org jstirnaman@influxdata.com
```

## What happened?

Output returned an error, even though `-i` switch was specified with a valid bucket ID:

```
Error: bucket missing: specify bucket ID or bucket name
```

## What is expected?

Listing the schema for a bucket succeeds:

```sh
07b81b87c6deb000        cpu                     4d5dbdfd5a67ae4f
07b80ed3bb73e000        temp                    4d5dbdfd5a67ae4f
```

stuartcarnie:
This PR ensures that either the `--bucket-id` or `-i` CLI options are permitted for the `bucket-schema` commands. It adds complete test coverage for all the `bucket-schema` commands to ensure the CLI options are permitted in all combinations.

Fixes #149

### DIFF
--- a/clients/bucket_schema/client.go
+++ b/clients/bucket_schema/client.go
@@ -49,10 +49,13 @@ func (c Client) resolveOrgBucketIds(ctx context.Context, params clients.OrgBucke
 	}
 
 	req := c.GetBuckets(ctx)
+	var nameID string
 	if params.BucketID.Valid() {
 		req = req.Id(params.BucketID.String())
+		nameID = params.BucketID.String()
 	} else {
 		req = req.Name(params.BucketName)
+		nameID = params.BucketName
 	}
 	if params.OrgID.Valid() {
 		req = req.OrgID(params.OrgID.String())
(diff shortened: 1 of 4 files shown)


clients/bucket_schema/client.go
```diff
@@ -40,19 +40,20 @@ func (c Client) resolveMeasurement(ctx context.Context, ids orgBucketID, name st
...
 	if !params.OrgID.Valid() && params.OrgName == "" && c.ActiveConfig.Org == "" {
 		return nil, errors.New("org missing: specify org ID or org name")
 	}
 
-	req := c.GetBuckets(ctx).Name(params.BucketName)
+	req := c.GetBuckets(ctx)
+	if params.BucketName != "" {
+		req = req.Name(params.BucketName)
+	} else {
+		req = req.Id(params.BucketID.String())
```
danxmoran, line 55: Do we want to prioritize the ID over the name in the case when both are set? That'd be consistent with how we manage org name/ID pairs. Not sure if it matters in practice.

danxmoran (approved):
One question, but overall LGTM. Thanks Stuart!

clients/bucket_schema/client.go
stuartcarnie: I'll switch them around if it is consistent with other commands 👍🏻 

Thanks for the review!

PR #152, opened 2021-06-25
feat: port `influx auth` command

### LINKED ISSUE #5: Port the `auth` command tree from `influxdb`
Port the `auth` command (and any subcommands) from the `influxdb` repo.

DStrand1:
Closes #5

No changes were needed for OpenAPI for this port (that I am aware of).


clients/auth/auth.go
```diff
@@ -0,0 +1,431 @@
...
+	}
+	return ret
+}
+
+func makePermResource(permType string, id string, orgId string, org string) api.PermissionResource {
+	return api.PermissionResource{
+		Type:  permType,
+		Id:    &id,
+		OrgID: &orgId,
+		Org:   &org,
```
danxmoran, line 429: The old CLI always uses `OrgID` (and looks up the ID if given the name), I think we need to follow that pattern

clients/auth/auth.go
```diff
@@ -0,0 +1,431 @@
...
+	ReadNotificationEndpointPermission  bool
+
+	WriteDBRPPermission bool
+	ReadDBRPPermission  bool
+}
+
+func (c Client) Create(ctx context.Context, params *CreateParams) error {
+
+	orgID := params.OrgID.String()
+	orgName := params.OrgName
```
danxmoran, line 81: Commented below, but I think we need to look up the org ID via the name if the ID isn't set

clients/auth/auth.go
```diff
@@ -0,0 +1,431 @@
...
+				Action:   bp.action,
+				Resource: makePermResource("buckets", p, orgID, orgName),
+			}
+			permissions = append(permissions, newPerm)
+		}
+	}
+
+	providedPerm := []struct {
+		readPerm, writePerm bool
+		ResourceType        string
```
danxmoran, line 108: Nitpick: Any reason to switch `lowerCase` and `UpperCase` in the field names here?

clients/auth/auth.go
```diff
@@ -0,0 +1,431 @@
...
+				Resource: makePermResource(provided.ResourceType, "", orgID, orgName),
+			}
+			permissions = append(permissions, p)
+		}
+	}
+
+	// Get the user ID because the command only takes a username, not ID
+	users, err := c.UsersApi.GetUsers(ctx).Name(params.User).Execute()
+	if err != nil || len(users.GetUsers()) == 0 {
+		return fmt.Errorf("could not find user with name %q: %q", params.User, err)
```
danxmoran, line 183:
```suggestion
		return fmt.Errorf("could not find user with name %q: %w", params.User, err)
```

clients/auth/auth.go
```diff
@@ -0,0 +1,431 @@
...
+	}
+	userID := users.GetUsers()[0].GetId()
+
+	authReq := api.AuthorizationPostRequest{
+		Description: &params.Description,
```
danxmoran, line 188: Can we update this to only set description if it's non-empty?

clients/auth/auth.go
```diff
@@ -0,0 +1,431 @@
...
+}
+
+func permString(p api.Permission) string {
+	ret := p.GetAction() + ":"
+	r := p.GetResource()
+
+	if r.GetOrgID() != "" {
+		ret += "orgs/" + r.GetOrgID()
+	}
+	ret += "/" + r.GetType()
```
danxmoran, line 417: If `GetOrgID()` returns empty string, we'll end up with a permission string like `action:/type`. I think all our permissions are org-scoped in practice at the moment, but it'd be nice to future-proof.

danxmoran: Also, I think this is copied from the v1 auth code. Can we centralize? You could create something like `api/permission_string.go` and then write it there as a method on `Permission`

danxmoran:
I just merged a PR that changes our codegen, you'll need to re-run `make openapi` to apply the changes to the newly-linked models

api/contract/cli.yml
```diff
@@ -369,3 +373,10 @@ components:
       $ref: "./openapi/src/common/schemas/DBRPUpdate.yml"
     DBRPGet:
       $ref: "./openapi/src/common/schemas/DBRPGet.yml"
+
+
```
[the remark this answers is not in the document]
DStrand1: I was hoping this would help avoid a merge conflict, but didn't seem to work out that way 😄

clients/auth/auth.go
DStrand1: Doing this would mean embedding OrganizationsApi into client for auth, do you think this is worth doing for this?

clients/auth/auth.go
DStrand1: Same as comment below

clients/auth/auth.go
danxmoran: Yes, creating via an org name isn't actually supported (I think the swagger API is being lazy and reusing the GET response for the POST body)

DStrand1: Okay, I will make this change now, otherwise this is ready for review again

PR #154, opened 2021-06-28
fix: downgrade urfave/cli to v1 to enable more flexible parsing

### LINKED ISSUE #147: Try downgrading to `urfave/cli` v1
`urfave/cli` is currently pretty strict about the ordering of flags & args on the CLI. This breaks some of our old CLI examples that pass flags after args. There's an upstream issue to relax the ordering constraint (urfave/cli#1113), but it sounds like some users have had success in just using v1 of the module (buildkite/agent#1286). Try downgrading and see what goes wrong.

danxmoran:
Closes #147

I've confirmed that rebasing #150 onto this branch fixes the breaking change (I can pass the stack ID before or after the `-f` flag).

### DIFF
--- a/clients/helpers.go
+++ b/clients/helpers.go
@@ -5,20 +5,17 @@ import (
 	"io/ioutil"
 	"os"
 	"strings"
-
-	"github.com/urfave/cli"
 )
 
 // ReadQuery reads a Flux query into memory from a --file argument, args, or stdin
-func ReadQuery(ctx *cli.Context) (string, error) {
-	nargs := ctx.NArg()
-	file := ctx.String("file")
+func ReadQuery(filepath string, args []string) (string, error) {
+	nargs := len(args)
 
 	if nargs > 1 {
-		return "", fmt.Errorf("at most 1 query string can be specified over the CLI, got %d", ctx.NArg())
+		return "", fmt.Errorf("at most 1 query string can be specified as an argument, got %d", nargs)
 	}
-	if nargs == 1 && file != "" {
-		return "", fmt.Errorf("query can be specified via --file or over the CLI, not both")
+	if nargs == 1 && filepath != "" {
+		return "", fmt.Errorf("query can be specified as a CLI arg or passed in a file via flag, not both")
 	}
 
 	readFile := func(path string) (string, error) {
@@ -37,14 +34,14 @@ func ReadQuery(ctx *cli.Context) (string, error) {
 		return string(queryBytes), err
 	}
 
-	if file != "" {
-		return readFile(file)
+	if filepath != "" {
+		return readFile(filepath)
 	}
 	if nargs == 0 {
 		return readStdin()
 	}
 
-	arg := ctx.Args().Get(0)
+	arg := args[0]
 	// Backwards compatibility.
 	if strings.HasPrefix(arg, "@") {
 		return readFile(arg[1:])
--- a/clients/helpers.go
+++ b/clients/helpers.go
@@ -6,7 +6,7 @@ import (
 	"os"
 	"strings"
 
-	"github.com/urfave/cli/v2"
+	"github.com/urfave/cli"
 )
 
 // ReadQuery reads a Flux query into memory from a --file argument, args, or stdin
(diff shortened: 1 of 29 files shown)


clients/helpers.go
```diff
@@ -5,20 +5,17 @@ import (
...
 	"strings"
-
-	"github.com/urfave/cli/v2"
 )
 
 // ReadQuery reads a Flux query into memory from a --file argument, args, or stdin
-func ReadQuery(ctx *cli.Context) (string, error) {
-	nargs := ctx.NArg()
-	file := ctx.String("file")
+func ReadQuery(filepath string, args []string) (string, error) {
```
danxmoran, line 11: The changes here are to allow kapacitor to upgrade `influx-cli` without needing to downgrade `urfave/cli` from v2 to v1. This was the only function that broke when I tried an upgrade locally.

wbaker85: Hmm can you help me out a bit on what this means? Since this PR is downgrading from v2 to v1 I'm not sure I follow. Is it because kapacitor uses `urfave/cli` v2?

cmd/influx/backup.go
```diff
@@ -6,15 +6,15 @@ import (
 	"github.com/influxdata/influx-cli/v2/clients/backup"
 	br "github.com/influxdata/influx-cli/v2/internal/backup_restore"
 	"github.com/influxdata/influx-cli/v2/pkg/cli/middleware"
-	"github.com/urfave/cli/v2"
+	"github.com/urfave/cli"
 )
 
-func newBackupCmd() *cli.Command {
```
wbaker85, line 12: Mostly out of curiosity...why did these all have to change from being pointers?

wbaker85: Guess it would help if I read the migration guide: https://github.com/urfave/cli/blob/master/docs/migrate-v1-to-v2.md#commands-are-now-lists-of-pointers

Although I do still wonder why that was 😃

clients/helpers.go
danxmoran: Kapacitor imports `influx-cli` as a dependency so it can reuse the business logic under `clients/` (including this function). It also implements its own CLI front-end for the tasks, using `urfave/cli/v2`. Downgrading the module in `influx-cli` changed the type of the `cli.Context` here, breaking the calls in Kapacitor. This change is still breaking, but it's easier to work around than the alternatives of either downgrading urfave or somehow using both versions there.

I'd been trying to keep any references to `urfave/cli` out of the `clients/` package to give us more flexibility in situations like this. This one slipped through the cracks, so I'm glad to refactor it away.

cmd/influx/backup.go
danxmoran: The types changed for some reason between V1->V2 of urfave. The migration guide is [here](https://github.com/urfave/cli/blob/master/docs/migrate-v1-to-v2.md), I effectively applied it in reverse.

wbaker85:
Everything looks good in the code as far as I can tell, but I have a higher level question as well - Is the issue with v2 breaking the old CLI examples really just related to examples - docs examples I'm guessing? I'm thinking along the lines of updating the examples vs. using an older version of `urfave`.

Updating a bunch of docs wouldn't be very much fun but that might have some pros instead of working with an older/not maintained package. If there's more significant breakages that would occur with the new behavior maybe it would outweigh that though.

danxmoran:
> Is the issue with v2 breaking the old CLI examples really just related to examples - docs examples I'm guessing?

Yes, across our public docs, help text, and (I'd guess) blog posts. I think it's something we could eventually phase out w/ proper deprecation notice, but I think it'd be a big lift to update everything in time for the new CLI release (esp. given the time pressure). I think minimizing the number of big incompatibilities will be important to get people to go through the effort of migrating to the new version; we're still paying for some big mismatches between v1 and v2, I don't want to add another layer of problems between v2.0 and v2.1.

I checked the release dates of `urfave/cli`. `v2.3.0` and `v1.22.5` were released within a day of each other, so I don't think we're regressing to a version that hasn't seen any maintenance.

wbaker85:
> > Is the issue with v2 breaking the old CLI examples really just related to examples - docs examples I'm guessing?
> 
> Yes, across our public docs, help text, and (I'd guess) blog posts. I think it's something we could eventually phase out w/ proper deprecation notice, but I think it'd be a big lift to update everything in time for the new CLI release (esp. given the time pressure). I think minimizing the number of big incompatibilities will be important to get people to go through the effort of migrating to the new version; we're still paying for some big mismatches between v1 and v2, I don't want to add another layer of problems between v2.0 and v2.1.
> 
> I checked the release dates of `urfave/cli`. `v2.3.0` and `v1.22.5` were released within a day of each other, so I don't think we're regressing to a version that hasn't seen any maintenance.

wbaker85 (approved):

PR #157, opened 2021-06-28
fix: correct mismatches between template-apply spec and impl

danxmoran:
Part of #4 

Experimenting with the dry-run piece of `apply` flushed out a few issues:
* Some collection fields needed to be marked as required for ergonomics (`map` is more usable than `*map`, and the pointer doesn't give us much)
* A missing required field in template application
* A misnamed field in template application
* Most significantly, all IDs returned by template application come back as large ints, not strings
   * I opened influxdata/openapi#158 to track the problem in the source-of-truth. For now, Greg and I decided to leave the server-side impls as they are and work around it in the CLI.
   * The ints that are returned from the server are random `unit64`s, so the usual trick of `type: integer` + `format: int64` doesn't work (I saw overflows when unmarshalling)
   * To fix the above, I introduced a new vendor extension for setting an explicit type. This required making changes to our codegen template

### DIFF
--- a/api/contract/overrides/schemas/TemplateApply.yml
+++ b/api/contract/overrides/schemas/TemplateApply.yml
@@ -14,7 +14,8 @@ properties:
       $ref: "./TemplateApplyTemplate.yml"
   envRefs:
     type: object
-    additionalProperties: true
+    additionalProperties:
+      type: string
   secrets:
     type: object
     additionalProperties:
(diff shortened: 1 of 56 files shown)


api/contract/overrides/schemas/TemplateApply.yml
```diff
@@ -14,7 +14,8 @@ properties:
       $ref: "./TemplateApplyTemplate.yml"
   envRefs:
     type: object
-    additionalProperties: true
```
danxmoran, line 17: This was generating the unfortunate type of `map[string]map[string]interface{}`

wbaker85 (approved):
Looks good!

PR #158, opened 2021-06-29
fix: fix extension-check logic when normalizing github URLs

danxmoran:
Part of #4

### DIFF
--- a/pkg/github/normalize.go
+++ b/pkg/github/normalize.go
@@ -37,7 +37,7 @@ func NormalizeURLToContent(u *url.URL, extensions ...string) *url.URL {
 func extensionMatches(u *url.URL, extensions []string) bool {
 	ext := path.Ext(u.Path)
 	for _, e := range extensions {
-		if strings.EqualFold(e, "."+ext) {
+		if strings.EqualFold(ext, "."+e) {
 			return true
 		}
 	}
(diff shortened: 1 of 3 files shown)


pkg/github/normalize.go
```diff
@@ -37,7 +37,7 @@ func NormalizeURLToContent(u *url.URL, extensions ...string) *url.URL {
 func extensionMatches(u *url.URL, extensions []string) bool {
 	ext := path.Ext(u.Path)
 	for _, e := range extensions {
-		if strings.EqualFold(e, "."+ext) {
+		if strings.EqualFold(ext, "."+e) {
```
wbaker85, line 40: A possible thought:

I see where the `NormalizeURLToContent` function is used here: https://github.com/influxdata/influx-cli/blob/main/clients/apply/source.go#L130-L145

In the switch statement, all of the extensions are listed with the `.` prefix. Then in the function call, they don't have the `.` prefix. I'm wondering if you could do the function call with the extensions having the dot prefix, like 

```
normalized := github.NormalizeURLToContent(u, ".yaml", ".yml", ".jsonnet", ".json").String()
```

Then you shouldn't need the `"." + e` here I don't think. It also strikes me as being a little more intuitive to have the strings in the cases for the switch statement match up 1:1 with extensions in the argument list.

This might be overkill and/or more effort to change now than what it's worth, but something to consider. Easy to nitpick when looking at a change to a single line of code, so feel free to take it or leave it!

wbaker85 (approved):

pkg/github/normalize.go
danxmoran: I'll take it!

PR #159, opened 2021-06-29
fix: correct more details in templates/apply API

danxmoran:
Part of #4 

With these changes, I was able to get `apply` fully ported.


api/contract/overrides/schemas/TemplateSummaryTelegrafConfig.yml
```diff
@@ -1,9 +1,7 @@
 type: object
 properties:
   id:
-    type: integer
-    format: int64
-    x-go-field-type: uint64
+    type: string
```
wbaker85, line 4: Hmmm why'd this one have to change back to a string?

danxmoran: The `pkger` code on the server-side declares custom models w/ a custom ID type for everything but telegrafs. For telegrafs, it embeds a copy of the `TelegrafConfig` model used elsewhere in the system. That reused model contains a "normal" `ID` field, which properly implements `MarshalText` and therefore gets stringified when the response is set. The custom ID type used in every other response doesn't implement `MarshalText`, so its instances get returned as unit64s...

`pkger` really needs a facelift

wbaker85 (approved):

PR #160, opened 2021-06-29
feat: port `apply` command from `influxdb`

### LINKED ISSUE #4: Port the `apply` command tree from `influxdb`
Port the `apply` command (and any subcommands) from the `influxdb` repo.

danxmoran:
Closes #4

The logic here isn't a 1:1 match. At the parsing layer, I changed:
* `--template-url / -u` is now un-deprecated
* Passing a HTTP(s) URL to `--file / -f` is now deprecated
* Malformed values passed to `--env-ref / --secret / --filter` now result in an error instead of being silently dropped

Within the core logic, the CLI code no longer knows how to summarize templates. Because of this:
* The CLI locates all missing env & secret refs by POST-ing a dry-run to the server and inspecting the returned summary
* To display an accurate diff, the CLI posts a 2nd dry-run after collecting all missing env/secret values, and uses the result of _that_ to render the diff table
* The 2nd dry-run is skipped if there are no missing values

### DIFF
--- a/clients/apply/apply.go
+++ b/clients/apply/apply.go
@@ -1,0 +1,729 @@
+package apply
+
+import (
+	"context"
+	"encoding/json"
+	"errors"
+	"fmt"
+	"reflect"
+	"strconv"
+	"strings"
+	"time"
+
+	"github.com/influxdata/influx-cli/v2/api"
+	"github.com/influxdata/influx-cli/v2/clients"
+	"github.com/influxdata/influx-cli/v2/pkg/influxid"
+	"github.com/influxdata/influx-cli/v2/pkg/template"
+)
+
+type Client struct {
+	clients.CLI
+	api.TemplatesApi
+	api.OrganizationsApi
+}
+
+type Params struct {
+	OrgId   string
+	OrgName string
+
+	StackId   string
+	Sources   []TemplateSource
+	Recursive bool
+
+	Secrets map[string]string
+	EnvVars map[string]string
+
+	Filters []ResourceFilter
+
+	Force              bool
+	OverwriteConflicts bool
+	Quiet              bool
+	RenderTableColors  bool
+	RenderTableBorders bool
+}
+
+type ResourceFilter struct {
+	Kind string
+	Name *string
+}
+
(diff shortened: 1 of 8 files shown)


clients/apply/apply.go
```diff
@@ -0,0 +1,729 @@
...
+	if !params.Quiet {
+		if err := c.printSummary(res.Summary, params); err != nil {
+			return err
+		}
+	}
+
+	return nil
+}
+
+func readTemplates(ctx context.Context, sources []TemplateSource) ([]api.TemplateApplyTemplate, error) {
```
danxmoran, line 161: I expect that some of these helper functions will move to separate files so they can be reused in the remaining template-related commands.

clients/apply/apply.go
```diff
@@ -0,0 +1,729 @@
...
+	}
+
+	// Initial dry-run to make the server summarize the template & report any missing env/secret values.
+	res, err := c.ApplyTemplate(ctx).TemplateApply(req).Execute()
+	if err != nil {
+		return fmt.Errorf("failed to check template impact: %w", err)
+	}
+
+	if len(res.Summary.MissingEnvRefs) > 0 || len(res.Summary.MissingSecrets) > 0 {
+		for _, e := range res.Summary.MissingEnvRefs {
```
wbaker85, line 104: There are some places in what looks like the equivalent code in `influxdb` where there is a check to see if whatever is happening is in a TTY or not, like here https://github.com/influxdata/influxdb/blob/deaf1dfe20aa39f26bf461aa9bd66283d0037238/cmd/influx/template.go#L246 - I noticed that wasn't always duplicated here, and am curious as to why not?

danxmoran: It felt redundant to the `--force` flag, which is what we use in other commands (i.e. `setup`) to bypass any interactive prompts. I guess it'll be a breaking change for any scripts out there that don't pass `--force`... will add in a check

danxmoran: Added

wbaker85 (approved):

Issue #142, opened 2021-06-23
Improve error messages even more for OSS- or cloud-only commands


wbaker85:
#140 added some basic functionality for flagging specific service APIs as being "OSS only" (ex: `health`, `backup`, `restore`) or "Cloud only" (`buckets-schemas`). When applied, these designations will expand on error messages if errors are encountered when executing the commands to indicate that the commands are OSS only or Cloud only as appropriate.

During the course of this work, it was noted that there could be some additional improvements:
- When https://github.com/influxdata/influxdb/issues/20224 & the equivalent cloud issue are implemented, use the `X-Influxdb-Version` header to explicitly verify that the server is appropriate for the command. A very specific error could be displayed if the command is OSS only, but the server response is from a cloud server, for example. It might even be possible to do a pre-flight check of some kind for commands like these to verify the right server is in place before executing the command, although that might limit backwards compatibility.
- #140 requires calling the "X-only" methods on individual services when they are created in the code. It might be better to have a separate list of of these special commands that could be maintained apart from the service creation logic. Although it wasn't immediately obvious at the time on how this could be done, it is an area for potential improvement in the future.

PR #165, opened 2021-06-30
fix: cloud or oss only error msg for parsable responses

### LINKED ISSUE #156: Cloud- or OSS- only commands which fail due to a server error do not have messages preserved
If a response comes back from the server and is able to be decoded successfully but is an error response for a Cloud- or OSS- only command, the error message will not have the appropriate additional text displayed as intended by #140.

This seems to be because if the code can make it all the way to this line: https://github.com/influxdata/influx-cli/blob/a058fe7e0b9f79eb1520f8184af1f7f3fd16ed8d/api/templates/api.mustache#L397

...the error returned is the `GenericOpenAPIError` which has its `model` property set to the decoded message. The `GenericOpenAPIError` implements the `error` interface by preferentially returning the result of the `error()` method on the `model` property rather than its own `error` property, see: https://github.com/influxdata/influx-cli/blob/main/api/templates/client.mustache#L570-L576

We need to fix this so that the `model` returns the additional information in the error message when `error()` is called on it.

wbaker85:
Closes #156

Some tweaks to the error message string formatting and return value for API call responses which are able to successfully be parsed (they are valid JSON), but are a status code outside the 200 range, indicating some kind of problem processing the request. This applies to API's flagged as Cloud- or OSS- only.

The error messages documented by #140 continue to be generated identically with this change, but this also adds the additional info text to errors such as trying to run a backup command against a cloud server which did not work as intended with #140:

```
influx backup .
2021/06/30 10:16:48 INFO: Downloading metadata snapshot
Error: failed to backup metadata: failed to download metadata snapshot: path not found
```

becomes:

```
influx backup .    
2021/06/30 10:15:53 INFO: Downloading metadata snapshot
Error: failed to backup metadata: failed to download metadata snapshot: InfluxDB OSS-only command failed
```

Having the additional text at the "end" of the error message isn't very satisfying, but I think it's the best we can do until we are able to do something like https://github.com/influxdata/influx-cli/issues/142

### DIFF
--- a/api/templates/api.mustache
+++ b/api/templates/api.mustache
@@ -376,7 +376,7 @@ func (a *{{{classname}}}Service) {{nickname}}Execute(r {{#structPrefix}}{{&class
 		}
 		newErr := GenericOpenAPIError{
 			body:  localVarBody,
-			error: _fmt.Sprintf("%s: code %s", errorPrefix, localVarHTTPResponse.Status),
+			error: _fmt.Sprintf("%s: %s", errorPrefix, localVarHTTPResponse.Status),
 		}
 		{{#responses}}
 		{{#dataType}}
@@ -388,10 +388,10 @@ func (a *{{{classname}}}Service) {{nickname}}Execute(r {{#structPrefix}}{{&class
 			var v {{{dataType}}}
 			err = a.client.decode(&v, localVarBody, localVarHTTPResponse.Header.Get("Content-Type"))
 			if err != nil {
-				newErr.error = _fmt.Sprintf("%s: %s", errorPrefix, err.Error())
+				newErr.error = _fmt.Sprintf("%s: %s", newErr.Error(), err.Error())
 				return {{#returnType}}localVarReturnValue, {{/returnType}}newErr
 			}
-			v.SetMessage(_fmt.Sprintf("%s: %s", errorPrefix, v.GetMessage()))
+			v.SetMessage(_fmt.Sprintf("%s: %s", newErr.Error(), v.GetMessage()))
 			newErr.model = &v
 			{{^-last}}
 			return {{#returnType}}localVarReturnValue, {{/returnType}}newErr
--- a/api/templates/api.mustache
+++ b/api/templates/api.mustache
@@ -391,7 +391,7 @@ func (a *{{{classname}}}Service) {{nickname}}Execute(r {{#structPrefix}}{{&class
 				newErr.error = _fmt.Sprintf("%s: %s", errorPrefix, err.Error())
 				return {{#returnType}}localVarReturnValue, {{/returnType}}newErr
 			}
-			v.SetMessage(errorPrefix)
+			v.SetMessage(_fmt.Sprintf("%s: %s", errorPrefix, v.GetMessage()))
 			newErr.model = &v
 			{{^-last}}
 			return {{#returnType}}localVarReturnValue, {{/returnType}}newErr
--- a/api/templates/api.mustache
+++ b/api/templates/api.mustache
@@ -358,25 +358,25 @@ func (a *{{{classname}}}Service) {{nickname}}Execute(r {{#structPrefix}}{{&class
 
 	var errorPrefix string
 	if a.isOnlyOSS {
-		errorPrefix = "InfluxDB OSS-only command failed: "
+		errorPrefix = "InfluxDB OSS-only command failed"
 	} else if a.isOnlyCloud {
-		errorPrefix = "InfluxDB Cloud-only command failed: "
+		errorPrefix = "InfluxDB Cloud-only command failed"
 	}
 
 	if localVarHTTPResponse.StatusCode >= 300 {
 		body, err := GunzipIfNeeded(localVarHTTPResponse)
 		if err != nil {
 			body.Close()
-			return {{#returnType}}localVarReturnValue, {{/returnType}}_fmt.Errorf("%s%w", errorPrefix, err)
+			return {{#returnType}}localVarReturnValue, {{/returnType}}_fmt.Errorf("%s: %w", errorPrefix, err)
 		}
 		localVarBody, err := _ioutil.ReadAll(body)
 		body.Close()
 		if err != nil {
-			return {{#returnType}}localVarReturnValue, {{/returnType}}_fmt.Errorf("%s%w", errorPrefix, err)
+			return {{#returnType}}localVarReturnValue, {{/returnType}}_fmt.Errorf("%s: %w", errorPrefix, err)
 		}
 		newErr := GenericOpenAPIError{
 			body:  localVarBody,
-			error: _fmt.Sprintf("%s%s", errorPrefix, localVarHTTPResponse.Status),
+			error: _fmt.Sprintf("%s: %s", errorPrefix, localVarHTTPResponse.Status),
 		}
 		{{#responses}}
 		{{#dataType}}
@@ -388,11 +388,11 @@ func (a *{{{classname}}}Service) {{nickname}}Execute(r {{#structPrefix}}{{&class
 			var v {{{dataType}}}
 			err = a.client.decode(&v, localVarBody, localVarHTTPResponse.Header.Get("Content-Type"))
 			if err != nil {
-				newErr.error = _fmt.Sprintf("%s%v", errorPrefix, err.Error())
+				newErr.error = _fmt.Sprintf("%s: %v", errorPrefix, err.Error())
 				return {{#returnType}}localVarReturnValue, {{/returnType}}newErr
 			}
+			v.SetMessage(errorPrefix)
 			newErr.model = &v
-			newErr.error = _fmt.Sprintf("%s%v", errorPrefix, v.Error())
 			{{^-last}}
 			return {{#returnType}}localVarReturnValue, {{/returnType}}newErr
 			{{/-last}}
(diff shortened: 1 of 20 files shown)


api/templates/api.mustache
```diff
@@ -358,25 +358,25 @@ func (a *{{{classname}}}Service) {{nickname}}Execute(r {{#structPrefix}}{{&class
...
 		localVarBody, err := _ioutil.ReadAll(body)
 		body.Close()
 		if err != nil {
-			return {{#returnType}}localVarReturnValue, {{/returnType}}_fmt.Errorf("%s%w", errorPrefix, err)
+			return {{#returnType}}localVarReturnValue, {{/returnType}}_fmt.Errorf("%s: %w", errorPrefix, err)
 		}
 		newErr := GenericOpenAPIError{
 			body:  localVarBody,
-			error: _fmt.Sprintf("%s%s", errorPrefix, localVarHTTPResponse.Status),
+			error: _fmt.Sprintf("%s: %s", errorPrefix, localVarHTTPResponse.Status),
```
danxmoran, line 379: Slightly more clear(?):
```suggestion
			error: _fmt.Sprintf("%s: code %s", errorPrefix, localVarHTTPResponse.Status),
```

api/templates/api.mustache
```diff
@@ -388,11 +388,11 @@ func (a *{{{classname}}}Service) {{nickname}}Execute(r {{#structPrefix}}{{&class
 			var v {{{dataType}}}
 			err = a.client.decode(&v, localVarBody, localVarHTTPResponse.Header.Get("Content-Type"))
 			if err != nil {
-				newErr.error = _fmt.Sprintf("%s%v", errorPrefix, err.Error())
+				newErr.error = _fmt.Sprintf("%s: %v", errorPrefix, err.Error())
```
danxmoran, line 391:
```suggestion
				newErr.error = _fmt.Sprintf("%s: %w", errorPrefix, err.Error())
```

api/templates/api.mustache
```diff
@@ -414,18 +414,18 @@ func (a *{{{classname}}}Service) {{nickname}}Execute(r {{#structPrefix}}{{&class
...
 	if err != nil {
-		return {{#returnType}}localVarReturnValue, {{/returnType}}_fmt.Errorf("%s%w", errorPrefix, err)
+		return {{#returnType}}localVarReturnValue, {{/returnType}}_fmt.Errorf("%s: %w", errorPrefix, err)
 	}
 	err = a.client.decode(&localVarReturnValue, localVarBody, localVarHTTPResponse.Header.Get("Content-Type"))
 	if err != nil {
 		newErr := GenericOpenAPIError{
 			body:  localVarBody,
-			error: _fmt.Sprintf("%s%s", errorPrefix, err.Error()),
+			error: _fmt.Sprintf("%s: %s", errorPrefix, err.Error()),
```
danxmoran, line 428: Hmmm why not this?
```suggestion
			error: _fmt.Sprintf("%s: %w", errorPrefix, err),
```

api/templates/api.mustache
```diff
@@ -388,11 +388,11 @@ func (a *{{{classname}}}Service) {{nickname}}Execute(r {{#structPrefix}}{{&class
 			var v {{{dataType}}}
 			err = a.client.decode(&v, localVarBody, localVarHTTPResponse.Header.Get("Content-Type"))
 			if err != nil {
-				newErr.error = _fmt.Sprintf("%s%v", errorPrefix, err.Error())
+				newErr.error = _fmt.Sprintf("%s: %v", errorPrefix, err.Error())
 				return {{#returnType}}localVarReturnValue, {{/returnType}}newErr
 			}
+			v.SetMessage(errorPrefix)
```
danxmoran, line 394: This is the change, right? And everything else is adapting to it?

danxmoran: I think it would be valuable to still include the next of the error model. i.e. to show:
```
InfluxDB OSS-only command failed: path not found
```
instead of just:
```
InfluxDB OSS-only command failed
```
Even if it doesn't help users, it'll help us triage bug reports

api/templates/api.mustache
wbaker85: I hadn't tried this...while it does compile, it gives a linting warning everywhere: `fmt.Sprintf call has error-wrapping directive %w, which is only supported for functions backed by fmt.Errorf`, so it doesn't seem ideal.

Maybe there's a way to do this with error wrapping and setting the `model` field for the `GenericOpenAPIError` ... I'll look into it.

danxmoran: Oh I completely missed that this is `Sprintf`, not `Errorf`. In that case `%s` is fine 👍

api/templates/api.mustache
danxmoran: Missed that this is `Sprintf`, sorry 🤦

api/templates/api.mustache
wbaker85: I fixed the logic so that the status code part of the error string chain gets preserved all the way through rather than being overwritten, and the resulting error looks like this:

```
2021/06/30 11:36:33 INFO: Downloading metadata snapshot
Error: failed to backup metadata: failed to download metadata snapshot: InfluxDB OSS-only command failed: 404 Not Found: path not found
````

I hadn't been able to see the status code in any error messages before, but now that I see that it includes the both the response code and the response text, having the extra text in front of that seems redundant.

File: api/templates/api.mustache (hunk 3 above)
Comment (wbaker85) [reply api/templates/api.mustache:428]: I changed the above one to `%s` as well so that all of them are consistent now. Not sure why it was a `%v`, must have just been a hold-over from before. Basically this entire chain of logic is just building up a string.

On that note, I looked at trying to do something more idiomatic like building up the error with `%w` and `fmt.Errorf` and didn't have much success. Untangling that from the error types in the template file here didn't seem to work very well and you'd end up converting back and forth between `error` and `string` all the time anyway. 

It would be nice to revisit this with https://github.com/influxdata/influx-cli/issues/142. I'm liking the idea more and more of a pre-flight request for server-specific commands and erroring out before even trying to run the command, and then just reverting all of this error building stuff back to the defaults from the template.

File: api/templates/api.mustache (hunk 4 above)
Comment (wbaker85) [reply api/templates/api.mustache]: Yep this is pretty much the fundamental change! I updated the code to preserve & display the actual error contained in `v` along with the prefix; good idea.

File: api/templates/api.mustache [hunk 5]
```diff
@@ -388,11 +388,11 @@ func (a *{{{classname}}}Service) {{nickname}}Execute(r {{#structPrefix}}{{&class
 			var v {{{dataType}}}
 			err = a.client.decode(&v, localVarBody, localVarHTTPResponse.Header.Get("Content-Type"))
 			if err != nil {
-				newErr.error = _fmt.Sprintf("%s%v", errorPrefix, err.Error())
+				newErr.error = _fmt.Sprintf("%s: %s", newErr.Error(), err.Error())
```
wbaker85, line 391: This change is to continue to build on the error message from `newErr` above; which contains the relevant error prefix (if any), and the response status like `404 Not Found`.

wbaker85: Also changed the `%v` to be a `%s` to be consistent with the `Sprintf`'s elsewhere, since we are dealing with strings.

api/templates/api.mustache
```diff
@@ -388,11 +388,11 @@ func (a *{{{classname}}}Service) {{nickname}}Execute(r {{#structPrefix}}{{&class
...
 				return {{#returnType}}localVarReturnValue, {{/returnType}}newErr
 			}
+			v.SetMessage(_fmt.Sprintf("%s: %s", newErr.Error(), v.GetMessage()))
```
wbaker85, line 394: This sets the "message" of `v` to include the previous error string in addition to the decoded error message from above.

api/templates/api.mustache
```diff
@@ -388,11 +388,11 @@ func (a *{{{classname}}}Service) {{nickname}}Execute(r {{#structPrefix}}{{&class
...
 			newErr.model = &v
-			newErr.error = _fmt.Sprintf("%s%v", errorPrefix, v.Error())
```
wbaker85, line 395: Getting rid of this line since it wasn't actually doing anything - the error message is returned from the `model` assigned to `v` on the line above.

danxmoran (approved):

wbaker85:
After everything's said and done, the error messages look like this...

For a general command that is not specific to cloud or OSS and encounters some kind of error on the server (no mention of cloud or OSS-only needed here):
```
influx v1 dbrp list --org invalid_org
Error: failed to list dbrps: 404 Not Found: invalid org "invalid_org": organization not found
```

For a command that is specific to OSS which is used with a cloud server config and the response cannot be parsed:
```
influx ping        
```

The reverse; a cloud-only command used with an OSS server:
```
influx bucket-schema list -n bucket -o org
Error: failed to list measurement schemas: InfluxDB Cloud-only command failed: 404 Not Found: unable to decode response content type "text/plain; charset=utf-8"
```

Finally, the original goal of this PR, a server specific command used on an incompatible backend with a resposne that IS able to be parsed:

```
influx backup .    
2021/06/30 12:31:58 INFO: Downloading metadata snapshot
Error: failed to backup metadata: failed to download metadata snapshot: InfluxDB OSS-only command failed: 404 Not Found: path not found
```

PR #163, opened 2021-06-30
feat: add stacks APIs to codegen

danxmoran:
Part of #18

Had to override everything again here to split up the `schemas` and tweak the tags on the routes. I made #162 to track revisiting all the overrides after we've pushed the initial release.

PR #166, opened 2021-06-30
chore: fix up openapi codegen after multi-merge

danxmoran:
Merging #165 and #163 in sequence broke our `openapi` CI check because I didn't re-generate in-between to pick up the new error logic. This re-generates everything, and also fixes the bad strings I missed in #165.


wbaker85 (approved):
👍

PR #168, opened 2021-06-30
feat: port `stacks` command from `influxdb`

### LINKED ISSUE #18: Port the `stacks` command tree from `influxdb`
Port the `stacks` command (and any subcommands) from the `influxdb` repo.

danxmoran:
Closes #18


wbaker85:
Everything looks good...one question, are you planning on writing unit tests for the client code at some point?

danxmoran:
> are you planning on writing unit tests for the client code at some point?

At some point, yes 😆 My opinion of the mock-based tests has gradually soured, especially for routes where we've overridden the entire API so our mocks might not be based on reality. I intend to investigate & set up a proper e2e testing suite once the pressure to port all the commands over has passed; there will be fewer tests overall, but I'm hopeful each test will be more valuable than a mocked-out unit test.

That said, there are some interesting helper functions in here that could be unit-tested. I'll add some!

wbaker85 (approved):

Issue #132, opened 2021-06-17
Cleanup duplicate logic for acquiring Organization ID


DStrand1:
There is a lot of duplicate logic for acquiring the organization ID from passed in `OrgParams`. There are two main situations that "business logic" for commands has to deal with this:
### Has an embedded `OrganizationsApi`:
This case is relatively simple, and can cleanup the logic with a method along the lines of:
```go
func (o OrgParams) GetOrgID(ctx context.Context, api api.OrganizationsApi, defaultOrg string) (string, error) {
    ...
}
```
where `defaultOrg` is intended to be passed in from the `ActiveConfig`, unless special logic demands it be different.

### Does not have an embedded `OrganizationsApi`:
My first thought for this was having an interface like:
```go
type <T> OrgHolder interface {
    Org(string) T
    OrgID(string) T
}

func (o OrgParams) <T> GetOrgID(defaultOrg string) T {
    ...
}
```
since other embedded api's that allow search by OrgName or OrgID will have these two API methods. However clearly we cannot do this, since Go currently lacks generics. This case is more difficult to solve, and will need some thought/discussion.

lesam:
Closed by https://github.com/influxdata/influx-cli/pull/272

PR #169, opened 2021-07-01
feat: implement `template` command

### LINKED ISSUE #21: Port the `template` command tree from `influxdb`
Port the `template` command (and any subcommands) from the `influxdb` repo.

danxmoran:
Closes #21

This isn't a 1:1 port from the code in `influxdb`:
* The old CLI performed summarization / validation completely on the client-side, reusing the code that the server runs when parsing `apply` requests
* This implementation POSTs a dry-run `apply` to the server and parses the response to replicate the behavior without needing to reimplement all the server logic in the new repo

I sanity-checked with @russorat before taking the new approach.

### DIFF
--- a/api/model_template_summary_variable_args_render.go
+++ b/api/model_template_summary_variable_args_render.go
@@ -3,10 +3,11 @@ package api
 import (
 	"encoding/json"
 	"fmt"
+	"log"
 	"strings"
 )
 
-func (args *TemplateSummaryVariableArgs) String() string {
+func (args *TemplateSummaryVariableArgs) Render() string {
 	if args == nil {
 		return "<nil>"
 	}
@@ -14,13 +15,15 @@ func (args *TemplateSummaryVariableArgs) String() string {
 	case "map":
 		b, err := json.Marshal(args.Values)
 		if err != nil {
-			return "{}"
+			log.Printf("WARN: failed to parse map-variable args: expected JSON, got %v\n", args.Values)
+			return "<parsing err>"
 		}
 		return string(b)
 	case "constant":
 		values, ok := args.Values.([]interface{})
 		if !ok {
-			return "[]"
+			log.Printf("WARN: failed to parse constant-variable args: expected array, got %v\n", args.Values)
+			return "<parsing err>"
 		}
 		var out []string
 		for _, v := range values {
@@ -30,18 +33,21 @@ func (args *TemplateSummaryVariableArgs) String() string {
 	case "query":
 		values, ok := args.Values.(map[string]interface{})
 		if !ok {
-			return ""
+			log.Printf("WARN: failed to parse query-variable args: expected JSON object, got %v\n", args.Values)
+			return "<parsing err>"
 		}
 		qVal, ok := values["query"]
 		if !ok {
-			return ""
+			log.Printf("WARN: failed to parse query-variable args: no 'query' key in %v\n", values)
+			return "<parsing err>"
 		}
 		lVal, ok := values["language"]
 		if !ok {
-			return ""
+			log.Printf("WARN: failed to parse query-variable args: no 'language' key in %v\n", values)
+			return "<parsing err>"
 		}
 		return fmt.Sprintf("language=%q query=%q", lVal, qVal)
 	default:
 	}
-	return "unknown variable argument"
+	return "<unknown variable type>"
 }
--- a/api/model_template_summary_variable_args_render.go
+++ b/api/model_template_summary_variable_args_render.go
@@ -1,0 +1,47 @@
+package api
+
+import (
+	"encoding/json"
+	"fmt"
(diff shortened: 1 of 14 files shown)


api/model_template_summary_variable_args_render.go
```diff
@@ -0,0 +1,47 @@
...
+	case "map":
+		b, err := json.Marshal(args.Values)
+		if err != nil {
+			return "{}"
+		}
+		return string(b)
+	case "constant":
+		values, ok := args.Values.([]interface{})
+		if !ok {
+			return "[]"
```
wbaker85, line 23: Hmm feels like you might want to log something for these to indicate that the type casting didn't work? Unless it's expected to be `!ok` normally? (That would be kind of strange..?)

clients/template/template.go
```diff
@@ -0,0 +1,119 @@
...
+	RenderTableColors  bool
+	RenderTableBorders bool
+}
+
+func (c Client) Summarize(ctx context.Context, params *SummarizeParams) error {
+	if params.OrgId == "" && params.OrgName == "" && c.ActiveConfig.Org == "" {
+		return clients.ErrMustSpecifyOrg
+	}
+	orgID := params.OrgId
+	if orgID == "" {
```
wbaker85, line 33: crossing fingers for https://github.com/influxdata/influx-cli/issues/132 🤞

pkg/template/summary_printer.go
```diff
@@ -0,0 +1,194 @@
...
+	"fmt"
+	"io"
+	"time"
+
+	"github.com/influxdata/influx-cli/v2/api"
+	"github.com/influxdata/influx-cli/v2/pkg/influxid"
+)
+
+func PrintSummary(summary api.TemplateSummaryResources, out io.Writer, useColor bool, useBorders bool) error {
+
```
wbaker85, line 13: nit: extra whitespace

pkg/template/summary_printer.go
```diff
@@ -0,0 +1,194 @@
...
+	newPrinter := func(title string, headers []string) *TablePrinter {
+		return NewTablePrinter(out, useColor, useBorders).
+			Title(title).
+			SetHeaders(append([]string{"Package Name", "ID", "Resource Name"}, headers...)...)
+	}
+
+	if labels := summary.Labels; len(labels) > 0 {
```
wbaker85, line 20: The thought has crossed my mind a few times while looking at code related to all this pkger stuff that "this sure is a lot of things that are almost the same but not quite".  To be sure, considering this is mostly a port of working code I don't think there is any reason to try to DRY it up right now, but if somebody gets bored one day...

clients/template/template.go
```diff
@@ -0,0 +1,119 @@
...
+		OrgID:     orgID,
+		Templates: templates,
+	}
+	_, err = c.ApplyTemplate(ctx).TemplateApply(req).Execute()
+	if err == nil {
+		return nil
+	}
+
+	if apiErr, ok := err.(api.GenericOpenAPIError); ok {
+		if summary, ok := apiErr.Model().(*api.TemplateSummary); ok {
```
wbaker85, line 114: Is this summary the same kind of summary as above on line 67? I'm wondering what it looks like when output here vs. from the `template.PrintSummary` output if so.

clients/template/template.go
```diff
@@ -0,0 +1,119 @@
...
+
+	// Execute a dry-run to make the server summarize the template(s).
+	req := api.TemplateApply{
+		DryRun:    true,
+		OrgID:     orgID,
+		Templates: templates,
+	}
+	res, err := c.ApplyTemplate(ctx).TemplateApply(req).Execute()
+	if err != nil {
+		return fmt.Errorf("failed to summarize template: %w", err)
```
wbaker85, line 61: On line 113 below, there's some additional processing in the `Validate` command that seems to still be getting a summary even if an error is returned. If that's what is actually happening, would it be useful to duplicate that kind of logic here? I haven't tried the command out myself, but I'm imagining some kind of useful summary info with maybe where the errors are - maybe it's nothing at all like that though and so it wouldn't be useful.

pkg/template/summary_printer.go
danxmoran: It's really begging for generics

clients/template/template.go
danxmoran: it is the same kind of summary. This command doesn't print anything for valid templates

clients/template/template.go
danxmoran: Tried this out, but looks like the server returns an empty summary when you dry-run something with validation errors.

danxmoran: influxdata/influxdb#21781

wbaker85 (approved):
Looks good!

PR #182, opened 2021-07-08
feat: update codegen to include legacy backup/restore routes

danxmoran:
All codegen (trying to keep the later diffs small)


wbaker85 (approved):
lgtm!

Issue #173, opened 2021-07-06
Extend `backup` to support backing up from OSS v2.0.x


danxmoran:
It should be possible for the new CLI to `backup` from OSS servers running v2.0.x. The CLI command will need to:
1. Detect the OSS version/capabilities using some API call
2. When detecting an "old" server, use `/backup/kv` instead of `/backup/metadata` to download the bolt file
3. Open up the downloaded bolt file to generate a new-style manifest
4. Run shard backups using the existing logic

PR #183, opened 2021-07-08
feat: add logic for reading manifest from bolt file

danxmoran:
Part of #173

This adds the minimal logic needed to open up a boltdb file downloaded from InfluxDB v2.0.x, extract the information relevant to backup/restore, and convert that info into a new-style manifest. It's not used by the `backup` command yet, but the included unit test shows it in action.


danxmoran:
Looks like Windows CI is running into etcd-io/bbolt#252...

wbaker85 (approved):
Looks good!

PR #184, opened 2021-07-09
feat: update `backup` to support InfluxDB 2.0.x

### LINKED ISSUE #173: Extend `backup` to support backing up from OSS v2.0.x

danxmoran:
Closes #173

### DIFF
--- a/clients/backup/backup.go
+++ b/clients/backup/backup.go
@@ -231,7 +231,12 @@ func (c *Client) downloadMetadata(ctx context.Context, params *Params) error {
 	return nil
 }
 
-// downloadMetadataLegacy TODO
+// downloadMetadataLegacy downloads a snapshot of the KV store from the server, and extracts
+// a bucket manifest from the result. KV is written to a local file; the extracted manifest
+// is tracked as a slice for additional processing.
+//
+// NOTE: This should _not_ be used against an InfluxDB instance running v2.1.0 or later, as
+// it will fail to capture metadata stored in SQL.
 func (c *Client) downloadMetadataLegacy(ctx context.Context, params *Params) error {
 	log.Println("INFO: Downloading legacy KV snapshot")
 	rawResp, err := c.GetBackupKV(ctx).Execute()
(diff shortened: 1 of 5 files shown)


clients/backup/backup.go
```diff
@@ -97,10 +145,12 @@ func (c *Client) downloadMetadata(ctx context.Context, params *Params) error {
...
 	if err != nil {
+		rawResp.Body.Close()
 		return err
 	}
 	body, err := api.GunzipIfNeeded(rawResp)
 	if err != nil {
+		rawResp.Body.Close()
 		return err
 	}
 	defer body.Close()
```
wbaker85, line 156: Can you put this `defer ...` on line ~142 and then remove the `rawResp.Body.Close()`'s on 148 and 153?

clients/restore/restore.go
```diff
@@ -146,13 +150,17 @@ func (c Client) fullRestore(ctx context.Context, path string) error {
...
 	}
-	log.Println("INFO: Restoring SQL snapshot")
-	if err := c.PostRestoreSQL(ctx).
-		ContentEncoding("gzip").
-		ContentType("application/octet-stream").
-		Body(sqlBytes).
-		Execute(); err != nil {
-		return fmt.Errorf("failed to restore SQL snapshot: %w", err)
+
+	// TODO: Should we have some way of wiping out any existing SQL on the server-side in the case when there is no backup?
```
wbaker85, line 154: This kind of situation seems like it would be a little messy, but I would tend to think it would be safer to not delete any existing data. If anything, we would probably want a Y/N prompt so the user doing the restore could decide if, when restoring a 2.0.x backup to a 2.1+ server, they want to wipe out existing sql data or not. It would require some server-side work as well...and it seems like it wouldn't come up very often. Something to keep in mind I guess if it happens more frequently than I am imagining.

clients/backup/backup.go
danxmoran: This defer couldn't move up because it has to be after `body` gets declared, but I changed the `rawResp.Body.Close()` calls to a single `defer` after the request. We might double-close the response but as far as I can tell that's no problem

wbaker85: nice!

wbaker85 (approved):
:+1:

PR #188, opened 2021-07-12
build: set up nightly build & release job

### LINKED ISSUE #170: Set up a nightly build & release
Similar to how things work in `influxdb`, every night we want to produce a release build & upload the result to our S3 bucket.

### LINKED ISSUE #187: Set up a CI `release` job
Similar to how things work in `influxdb`, when a git tag matching a semver pattern is pushed, we want to produce a release build & upload the result to our S3 bucket.

danxmoran:
Closes #170
Closes #187

Still need confirmation on:

- [x] Is the new naming pattern OK? Or do we want to reuse the old pattern from `influxdb`?
- [x] Do we want to sign artifacts from within CircleCI?
- [ ] How can we get the environment variables we need into our Circle env for this job?


.circleci/config.yml
```diff
@@ -89,6 +113,15 @@ jobs:
       - store_artifacts:
           path: dist
 
+  release:
+    executor: linux-amd64
+    steps:
+      - checkout
+      - setup
+      - run: make release
```
danxmoran, line 121: TODO: Inject `SNAPSHOT_VERSION` here, something like `nightly-<date>`

.circleci/config.yml
```diff
@@ -37,6 +37,71 @@ workflows:
...
+              ignore: /.*/
+            tags:
+              only: /^v[0-9]+\.[0-9]+\.[0-9]+(-(rc|alpha|beta)\.[0-9]+)?$/
+      - check-openapi:
+          filters:
+            branches:
+              ignore: /.*/
+            tags:
+              only: /^v[0-9]+\.[0-9]+\.[0-9]+(-(rc|alpha|beta)\.[0-9]+)?$/
+      - test:
```
codyshepherd, line 80: It might be worth it to have `test` require `lint`? That way we aren't wasting cycles testing if linting fails?

.goreleaser.yml
```diff
@@ -17,9 +22,57 @@ builds:
...
       - CGO_ENABLED=0
     ldflags:
-      - -s -w -X main.version={{if .IsSnapshot}}dev{{else}}{{.Version}}{{end}} -X main.commit={{.ShortCommit}} -X main.date={{.Date}}
+      - -s -w -X main.version={{if .IsSnapshot}}{{.Env.SNAPSHOT_VERSION}}{{else}}{{.Version}}{{end}} -X main.commit={{.ShortCommit}} -X main.date={{.Date}}
     binary: influx
 
-# Do not make github release
-release:
-  disable: true
+# TODO: Do we want to publish deb/rpms for the CLI before we release 2.1.0 of the server?
```
codyshepherd, line 28: Has Product weighed in on this yet? My feeling is that if we are developing and building the CLI separately, we should make packages available, but also this would require some nontrivial documentation of what the user path is and what we are prepared to support in terms of conflicts.

codyshepherd (changes requested):
A couple of inline notes. As to your questions:

* I think we should stick to the old naming pattern unless there's a good reason to change. AFAICT the new naming convention just substitutes an underscore for a dash in a couple places, is that right? If that's true, I'm not sure I see the justification for breaking with the old convention.
* I believe our path forward on signing in Circle is to hook it up to our Vault service and use that to do signing. I haven't yet looked into what that will take.
* Adding env variables to circle is just a matter of adjusting the "Project Settings" in the Circle UI. LMK if you don't have access and I can help with that

.circleci/config.yml
```diff
@@ -37,6 +37,71 @@ workflows:
...
+            - check-openapi
+            - test
+  release:
+    jobs:
+      - lint:
+          filters:
+            branches:
+              ignore: /.*/
+            tags:
+              only: /^v[0-9]+\.[0-9]+\.[0-9]+(-(rc|alpha|beta)\.[0-9]+)?$/
```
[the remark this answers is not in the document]
danxmoran: I think your regex-fu is correct. I copied this from `influxdb` 2.x but I've never actually used it to cut an rc/beta there. I'll update the pattern to match what we do in 1.x

.goreleaser.yml
danxmoran: Not yet, will ping Russ

.goreleaser.yml
```diff
@@ -17,9 +22,57 @@ builds:
...
+      - deb
+      - rpm
+    bindir: /usr/bin
+    overrides:
+      rpm:
+        replacements:
+          amd64: x86_64
+          arm64: aarch64
+          armhf: armv7hl
+        file_name_template: influx-cli-nightly.{{ .Arch }}
```
danxmoran, line 42: Self review: for all the name templates here, need to template out the version instead of hard-coding "nightly"

.goreleaser.yml
```diff
@@ -17,9 +22,57 @@ builds:
...
+  - signature: "${artifact}.asc"
+    cmd: gpg
+    args: [ --passphrase, "{{.Env.PASSPHRASE}}", --pinentry-mode=loopback, --batch, --armor, --detach-sign, "${artifact}"]
+    artifacts: all
+
+blobs:
+  - provider: s3
+    bucket: dl.influxdata.com
+    region: us-east-1
+    folder: platform/nightlies/
```
danxmoran, line 78: Self review: need to template this out.

danxmoran:
Re: the naming convention, the only existing upload we're replacing/superseding is the client-only archive, which used to be named:
```
influxdb2-client-{{ .Version }}-{{ .Os }}-{{ .Arch }}.[tar.gz|zip]
```
The client-only deb & rpm will be brand-new. The new CLI is compatible with the old CLI, so I think it should be ok to keep the same naming pattern.

The only reason I can think that might motivate us to change the pattern is if `influxdb` switches to something new (because all if its artifacts will now be server-only instead of server+client), and we want to make the CLI match/mirror the new thing.

.goreleaser.yml
danxmoran: Product aside, I was worried about releasing a 2.1 deb/rpm into the wild when the new package contains a file that's already present in `influxdb2` (`/usr/bin/influx`). It looks like the deb system has a way of specifying the breakage in this case: https://www.debian.org/doc/debian-policy/ch-relationships.html#s-replaces. Need to find if rpm has a similar concept

danxmoran: RPM doesn't have a 1:1 concept. What I found is:
* `yum install` doesn't work when there's a shared file between the two packages
* `rpm -i --replacefiles` does work as long as you install the CLI after the server

The work-around should only be needed until we release a version of `influxdb2` that only contains the server.

.circleci/config.yml
```diff
@@ -37,6 +37,71 @@ workflows:
...
                 - mac
                 - windows
+  release-nightly:
+    triggers:
+      - schedule:
+          cron: "30 20 * * *"
+          filters:
+            branches:
+              only:
+                - dm-nightly-build-170
```
danxmoran, line 47: Self-review: remember to revert these changes to the normal nightly cadence

.goreleaser.yml
codyshepherd: Interesting. yeah it does seem a bit messy, I wonder if we could release it initially as a series of rc releases to make it clear that it's not yet "GA" until we get the conflict worked out

Makefile
```diff
@@ -4,16 +4,31 @@ export GOOS=$(shell go env GOOS)
...
+else
+ifndef SNAPSHOT_VERSION
+	SNAPSHOT_VERSION := dev
+endif
+ifndef SNAPSHOT_PACKAGE_TAG
+	SNAPSHOT_PACKAGE_TAG := dev
+endif
+	LDFLAGS += -X main.version=$(SNAPSHOT_VERSION)
+	GORELEASER_SETENV := SNAPSHOT_VERSION=$(SNAPSHOT_VERSION) SNAPSHOT_PACKAGE_TAG=$(SNAPSHOT_PACKAGE_TAG)
+	SNAPSHOT_FLAG := --snapshot
```
danxmoran, line 31: Irritating, it seems `--snapshot` disables uploading to S3, so we can't use it for nightlies... investigating alternatives

.goreleaser.yml
danxmoran: For now I've added a "conflicts" clause to the RPMs. That'll prevent anyone from installing `influx-cli` if they already have `influxdb2`, which is annoying but hopefully not as annoying as installing `influx-cli` and discovering that the `influxd` binary has disappeared.

danxmoran:
@codyshepherd apart from reverting the test-cron, how does this look now?

codyshepherd (approved):

PR #191, opened 2021-07-13
feat: allow passing password via flag in `user password`

### LINKED ISSUE #88: should be able to change password for user from cli without interactive prompt
__Proposal:__
When changing password for user we should be able to specify new password  through flags instead of interactive prompt
influx user password -n <username> -p <new-password>

**I am using influxdb version v2.0.6 inside container**

__Current behavior:__
At present when we execute below command interactive terminal opens and asks for password and asks again for confirming same,it doesnt accept inputs from stdin
influx user password -n <username>

i tried below but it doesnt work
echo "abcd1234\nabcd1234\n" | influx user password -n influxdb

__Desired behavior:__
There should be option to set password by passing password though flags in command along with interactive shell as per user choice

__Alternatives considered:__
Describe other solutions or features you considered.

__Use case:__
Why is this important (helps with prioritizing requests)?
People who have automation scripts wont be able to change password for user they cant input values to interactive prompt

danxmoran:
Closes #88

### DIFF
--- a/clients/user/user.go
+++ b/clients/user/user.go
@@ -138,11 +138,12 @@ func (c Client) Update(ctx context.Context, params *UpdateParmas) error {
 }
 
 type SetPasswordParams struct {
-	Id   influxid.ID
-	Name string
+	Id       influxid.ID
+	Name     string
+	Password string
 }
 
-func (c Client) SetPassword(ctx context.Context, params *SetPasswordParams) error {
+func (c Client) SetPassword(ctx context.Context, params *SetPasswordParams) (err error) {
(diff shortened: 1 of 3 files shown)


wbaker85:
Looks like a nice addition & the code looks good! One question: How are additions like this going to be rolled into doc updates? Like the ones at https://docs.influxdata.com/influxdb/v2.0/reference/cli/

danxmoran:
I'm not sure how the docs team will want to message the split between CLI and server, should probably give them a heads-up about it...

As a starting point, I've got a CHANGELOG running in #188

wbaker85 (approved):

PR #192, opened 2021-07-13
refactor: expose method for building onboarding requests

danxmoran:
In support of influxdata/influxdb#21838, also make the `OnboardingRequest` method public so it can be reused.

### DIFF
--- a/clients/setup/setup.go
+++ b/clients/setup/setup.go
@@ -55,7 +55,7 @@ func (c Client) Setup(ctx context.Context, params *Params) error {
 	}
 
 	// Initialize the server.
-	setupBody, err := c.onboardingRequest(params)
+	setupBody, err := c.OnboardingRequest(params)
 	if err != nil {
 		return err
 	}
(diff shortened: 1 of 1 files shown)


clients/setup/setup.go
```diff
@@ -151,6 +152,25 @@ func (c Client) onboardingRequest(params *Params) (req api.OnboardingRequest, er
 	req.RetentionPeriodSeconds = &rpSecs
 
 	if params.Force {
+		// The server doesn't give a very helpful error message when required fields are missing,
```
wbaker85, line 155: Would it be practical at all to update the server code so that it gives better error messages in response rather than duplicating the check?

danxmoran: Looks like it'd be easy enough to apply the same fix on the server side. Interestingly, it looks like the password isn't actually required by the API even though the CLI has historically forced one to be set...

danxmoran: I'll back this out for now so we can merge the 1st commit, which is needed for `influxd upgrade` to work in PowerShell

wbaker85 (approved):

PR #195, opened 2021-07-13
fix: set content-encoding: gzip for non-legacy KV restore

danxmoran:
And this is why automated tests are a good thing...


wbaker85 (approved):
:+1:

PR #198, opened 2021-07-19
build: add GPG private key to gitignore

danxmoran:
This should fix the failure I hit when trying to push an RC0 (see [here](https://app.circleci.com/pipelines/github/influxdata/influx-cli/491/workflows/6074b67d-db3d-48ca-a3cf-1d0f602ae6e2/jobs/2738)). `influxdb` has the same entry in its gitignore: https://github.com/influxdata/influxdb/blob/master/.gitignore#L19


davidby-influx (approved):
LGTM

PR #202, opened 2021-07-21
refactor: move duration package out of internal

danxmoran:
Mechanical refactor, moving `internal/duration` to `pkg/duration` so its contents can be used by other projects.


lesam:
Seems like we should have a low-level repo to depend on rather than pulling the whole CLI for a dependency like this, but this is fine.

lesam (approved):

danxmoran:
> Seems like we should have a low-level repo to depend on rather than pulling the whole CLI for a dependency like this, but this is fine.

We're already pulling the CLI in `influxdb` for more than this, but agreed there seems to be room for a core/utils repo

PR #203, opened 2021-07-21
fix: retry password-prompt if 2nd entry doesn't match the 1st

### LINKED ISSUE #201: Password confirmation doesn't retry if 2nd entry is different from the 1st
Screen Shot 2021-07-21 at 11 25 09 AM

danxmoran:
Closes #201

New behavior in action:
Screen Shot 2021-07-21 at 11 58 17 AM

### DIFF
--- a/pkg/stdio/console.go
+++ b/pkg/stdio/console.go
@@ -87,19 +87,23 @@ func (t *terminalStdio) GetSecret(prompt string, minLen int) (password string, e
 // GetPassword prompts the user for a secret twice, and inputs must match.
 // Uses stdio.MinPasswordLen as the minimum input length
 func (t *terminalStdio) GetPassword(prompt string) (string, error) {
-	pass1, err := t.GetSecret(prompt, MinPasswordLen)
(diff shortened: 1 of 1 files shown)


pkg/stdio/console.go
```diff
@@ -87,19 +87,23 @@ func (t *terminalStdio) GetSecret(prompt string, minLen int) (password string, e
 // GetPassword prompts the user for a secret twice, and inputs must match.
 // Uses stdio.MinPasswordLen as the minimum input length
 func (t *terminalStdio) GetPassword(prompt string) (string, error) {
-	pass1, err := t.GetSecret(prompt, MinPasswordLen)
-	if err != nil {
-		return "", err
+	for {
+		pass1, err := t.GetSecret(prompt, MinPasswordLen)
```
danxmoran, line 91: This call will already retry internally if the user inputs a value that's too short

wbaker85 (approved):
Looks good... I tested this out on my end and it works correctly for me as well.

PR #208, opened 2021-07-27
feat: Bind skip-verify flag to INFLUX_SKIP_VERIFY

stuartcarnie:
This PR binds the `--skip-verify` flag to the `INFLUX_SKIP_VERIFY` environment variable. This allows a user to omit specifying the flag for every invocation of the `influx` command, and is particularly useful when using the `influx` command in development environments.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -32,13 +32,15 @@ The command now uses lowerCamelCase consistently for all objects keys, matching 
 ### Features
 
 1. [33](https://github.com/influxdata/influx-cli/pull/33): Add global `--http-debug` flag to help inspect communication with InfluxDB servers.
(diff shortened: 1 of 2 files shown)


danxmoran (approved):
Can you add a line to `CHANGELOG.md` as well?

russorat:
could/should we add this to the config profiles?

danxmoran:
> could/should we add this to the config profiles?

I think it would make sense for skip-verify to be in the profiles, but it'd touch enough code that I'd recommend a separate issue & PR for it.

stuartcarnie:
Will add to CHANGELOG.md and merge – thanks for the review, @danxmoran 🙇🏻

PR #213, opened 2021-07-29
test: remove go-cmp dependency

danxmoran:
I was double-checking our dependency list before pushing an official `v2.1.0`, and noticed that we added a dep on `go-cmp`. It was only used in 2 tests and could be easily replaced by uses of `testify`, so I dropped it to slim things down.


wbaker85 (approved):
👍

PR #219, opened 2021-07-30
build: prefix nightly versions with latest tag

### LINKED ISSUE #217: Update nightly versions to be prefixed by latest minor release
Our nightlies should be versioned like `2.<minor>.<patch>-nightly-<date>`. This will give more information to users, and help integrate our nightlies into systems like deb/rpm which rely on semver.

danxmoran:
Closes #217

### DIFF
--- a/.circleci/config.yml
+++ b/.circleci/config.yml
@@ -181,10 +181,12 @@ jobs:
           steps:
             - run:
                 name: Make a pseudo-tag for the nightly build
-                command: git tag nightly-$(date +%Y-%m-%d)
+                command: |
+                  latest_tag=$(git describe --tags --abbrev=0)
+                  git tag ${latest_tag}+nightly-$(date +%Y-%m-%d)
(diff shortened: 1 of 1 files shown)


codyshepherd (approved):

.circleci/config.yml
```diff
@@ -181,10 +181,12 @@ jobs:
           steps:
             - run:
                 name: Make a pseudo-tag for the nightly build
-                command: git tag nightly-$(date +%Y-%m-%d)
+                command: |
+                  latest_tag=$(git describe --tags --abbrev=0)
+                  git tag ${latest_tag}+nightly-$(date +%Y-%m-%d)
```
danxmoran, line 186: Based on [this document](http://www.fifi.org/doc/debian-policy/policy.html/ch-versions.html) I should probably move away from using `-` as the date separator, as the last one will get interpreted as the debian revision.

PR #236, opened 2021-08-20
fix: detect and error out on incorrect positional args

### LINKED ISSUE #235: `stacks` command doesn't show an error on incorrect subcommands
For example, this...
```
$ influx stacks foo
ID      OrgID   Active  Name    Description     Num Resources   Sources URLs    Created At      Updated At
```
... should instead print something like...
```
$ influx stacks foo
No help topic for 'foo'
```
.. with an error return code, to mimic the behavior of other commands.

This is probably a problem for all commands that define an `Action`, since we don't auto-enforce that no positional args are passed.

danxmoran:
Closes #235


davidby-influx (approved):
Seems straightforward

PR #251, opened 2021-09-01
build: upgrade to go1.17

### LINKED ISSUE #250: Upgrade to go1.17
1. Run `go mod tidy -go=1.17` to upgrade the contents of `go.mod`
   * If this creates a 2nd `require` block, combine the two and re-run the tidy command to re-sort the entries
2. Run `make fmt` to inject any needed `go:build` directives 
3. Make sure everything still builds & tests properly locally
4. Update CircleCI to use 1.17

danxmoran:
Closes #250

### DIFF
--- a/pkg/signals/context_test.go
+++ b/pkg/signals/context_test.go
@@ -1,4 +1,5 @@
-//+build !windows
+//go:build !windows
+// +build !windows
 
 package signals
 
--- a/scripts/ci/install-go.sh
+++ b/scripts/ci/install-go.sh
@@ -1,11 +1,31 @@
 #!/usr/bin/env bash
 set -eo pipefail
 
-declare -r GO_VERSION=1.16.5
+declare -r GO_VERSION=1.17
+
+# Hashes are from the table at https://golang.org/dl/
+function go_hash () {
+  case $1 in
+    linux_amd64)
+      echo 6bf89fc4f5ad763871cf7eac80a2d594492de7a818303283f1366a7f6a30372d
+      ;;
+    linux_arm64)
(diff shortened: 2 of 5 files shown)


pkg/signals/context_test.go
```diff
@@ -1,4 +1,5 @@
-//+build !windows
+//go:build !windows
```
davidby-influx, line 1: Do we want to move entirely to the new directive now, or wait until `+build` is deprecated?

scripts/ci/install-go.sh
```diff
@@ -1,25 +1,47 @@
 #!/usr/bin/env bash
 set -eo pipefail
 
-declare -r GO_VERSION=1.16.5
+declare -r GO_VERSION=1.17
+
+# Hashes are from the table at https://golang.org/dl/
```
davidby-influx, line 6: Any easy way to automate pulling the hashes directly? Or too much trouble...

davidby-influx (approved):
Non-blocking questions for my own edification only.

scripts/ci/install-go.sh
danxmoran: I think we'd need to parse the HTML of the page, so probably too much trouble :) having them hard-coded here is (IMO) part of the point anyways, so we don't have to worry about having our request to pull the hashes getting corrupted / intercepted.

pkg/signals/context_test.go
danxmoran: I'll see if `gofmt` allows for just the new directive, my impression from reading the release notes was that it'd enforce having both styles 👀

danxmoran: Didn't even get as far as running the build... GoLand immediately flags the `//go:build` with a "//go:build comment without //+build comment" error if I delete the `// +build` line

PR #252, opened 2021-09-01
fix: upgrade golang.org/x/sys to avoid panics on Macs

danxmoran:
When upgrading kapacitor to go1.17 I saw a panic from `go-isatty`, and tracked it down to `golang.org/x/sys`. golang/go#47279 says the bug has been fixed, so upgrading should prevent future panics.

I couldn't trigger the panic in the CLI but it does use `go-isatty`, so upgrading feels like a smart move.


davidby-influx (approved):
Good to go.

PR #254, opened 2021-09-02
feat: stub out `remote` and `replication` commands and subcommands

### LINKED ISSUE #239: Stub out `remotes` command and subcommands
Set up basic cobra command structure for `remotes` command and subcommands, including:

- `list`
- `create`
- `update` 
- `delete`

These commands will not yet be functional, but will output a response that that the command has been called correctly. Basic help text (-h or --help) will be implemented as part of this issue.

mcfarlm3:
Closes #239 and #240. Provides basic cobra command structure for two new top-level commands (`remote` and `replication`), as well as their corresponding subcommands. Functionality is not implemented yet. Top-level commands are currently marked as hidden, while subcommands are not.


cmd/influx/remote.go
```diff
@@ -0,0 +1,79 @@
...
+
+	"github.com/influxdata/influx-cli/v2/pkg/cli/middleware"
+	"github.com/urfave/cli"
+)
+
+func newRemoteCmd() cli.Command {
+	return cli.Command{
+		Name:   "remote",
+		Usage:  "Remote connection management commands",
+		Hidden: true, // Remove this line when all subcommands are completed
```
danxmoran, line 14: Can you make an issue for this within the epic so we don't forget to do it before releasing?

mcfarlm3: Yep, I'll do that now!

cmd/influx/replication.go
```diff
@@ -0,0 +1,94 @@
...
+		),
+		Action: func(ctx *cli.Context) {
+			fmt.Println("replication list command was called")
+		},
+	}
+}
+
+func newReplicationMetricsCmd() cli.Command {
+	return cli.Command{
+		Name:   "metrics",
```
danxmoran, line 70: Hmmm as I understand it, the only queue info we'll have for each stream is:
* Current # of bytes in the queue
* Max # of bytes in the queue

I'm not sure we need a separate command to show that vs. showing it in the output of `list`. What do you think?

mcfarlm3: I think it's fine to combine if that's all the info we need to show - I'll go ahead and change to reflect this.

danxmoran (approved):

PR #255, opened 2021-09-02
fix: --host value passed to setup is written to generated config

### LINKED ISSUE #253: influx setup doesn't set host correctly in the configs file for https hosts
to repro, start an influxd instance with tls configured, then try to set it up with the cli. once you do, check the `configs` file and notice that the host for the configs is configured to be `http://` instead of `https://`.

```
(base) ➜  influx-cli git:(main) ✗ git log | head -n 1
commit 38040fa8dcb1225b5d140cbe868583bd07e4382a

(base) ➜  influx-cli git:(main) ✗ ./bin/darwin/influx version                                                                                                                                         
Influx CLI dev (git: none) build_date: 2021-09-02T16:01:07Z

(base) ➜  influx-cli git:(main) ✗ ./bin/darwin/influx setup --host https://$(docker-machine ip):8086 --skip-verify --configs-path ./configs -u russ -p something -t mytoken -o InfluxData -b telegraf -r 7d -f
User    Organization    Bucket
russ    InfluxData      telegraf

(base) ➜  influx-cli git:(main) ✗ cat configs 
[default]
  url = "http://localhost:8086"
  token = "mytoken"
  org = "InfluxData"
```

The expectation is that the configs file should use the `--host` value in the default config.

danxmoran:
Closes #253

The previous behavior would wrongly copy the host of the currently-active config to the newly-generated config.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -10,6 +10,7 @@ This release upgrades the project to `go` version 1.17.
 1. [228](https://github.com/influxdata/influx-cli/pull/228): Make global `--http-debug` flag visible in help text.
 1. [232](https://github.com/influxdata/influx-cli/pull/232): Don't set empty strings for IDs in permission resources.
 1. [236](https://github.com/influxdata/influx-cli/pull/236): Detect and error out on incorrect positional args.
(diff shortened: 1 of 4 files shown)


wbaker85 (approved):
LGTM!

PR #259, opened 2021-09-08
chore: add shorthand for `--bucket` to `influx delete`

DStrand1:
When trying to test my issue with `SHOW TAG VALUES`, I noticed that most commands (for example, `influx write`) allow for specifying either `--bucket` or `-b`, however `delete` only allowed `--bucket`. This PR adds the ability to use `-b` for `delete`

### DIFF
--- a/cmd/influx/delete.go
+++ b/cmd/influx/delete.go
@@ -33,7 +33,7 @@ func newDeleteCmd() cli.Command {
 				Value:  &params.BucketID,
 			},
 			&cli.StringFlag{
-				Name:        "bucket",
+				Name:        "bucket, b",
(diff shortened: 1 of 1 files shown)


cmd/influx/delete.go
```diff
@@ -33,7 +33,7 @@ func newDeleteCmd() cli.Command {
 				Value:  &params.BucketID,
 			},
 			&cli.StringFlag{
-				Name:        "bucket",
+				Name:        "bucket, b",
```
danxmoran, line 36: Can you add a note to the CHANGELOG.md for this? We haven't automated it here yet

danxmoran (approved):

PR #267, opened 2021-09-10
chore(Makefile): allow specifying -gcflags via env

jdstrand:
Make targets may call '$(GO_BUILD)', but there is no facility for
specifying arguments to 'go build'. As a first step, introduce the
GCFLAGS Makefile variable that when unset, operates as always, but when
set, adds '-gcflags "$(GCFLAGS)"' to go build. Eg, when unspecified,
maintain the current behavior (though with an additional space):
```
$ make
CGO_ENABLED=0 go build  -ldflags ...
```

When specified, add the specified -gcflags:
```
$ GCFLAGS="all=-N -l" make
CGO_ENABLED=0 go build -gcflags "all=-N -l" -ldflags ...
```
This could be useful in various situations such as producing unoptimized
builds (like in the above).

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -12,7 +12,15 @@ endif
 ifdef COMMIT
 	LDFLAGS += -X main.commit=$(COMMIT)
 endif
-export GO_BUILD=go build -ldflags "$(LDFLAGS)"
+
+# Use default flags, but allow adding -gcflags "..." if desired. Eg, for debug
+# builds, may want to use GCFLAGS="all=-N -l" in the build environment.
+GCFLAGS ?=
+ifneq ($(GCFLAGS),)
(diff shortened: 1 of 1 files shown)


jdstrand:
For some reason, the CI tests didn't run and I don't seem to have the ability to poke them along.

danxmoran:
@jdstrand I had to toggle a setting in CircleCI to enable building PRs from forks. It should work now (but you might need to push a rebase / empty commit to trigger it)

jdstrand:
> @jdstrand I had to toggle a setting in CircleCI to enable building PRs from forks. It should work now (but you might need to push a rebase / empty commit to trigger it)

Things seem to be running now (I did `git commit --amend`, didn't change anything and force pushed). Thanks! :)

PR #268, opened 2021-09-14
feat: added functionality for `remote create` subcommand

### LINKED ISSUE #242: Implement `remotes create` subcommand
This command should register/create a new remote connection

mcfarlm3:
Closes #242 

This command creates/registers a new remote connection.


clients/remote/remote.go
```diff
@@ -0,0 +1,97 @@
...
+
+type Client struct {
+	clients.CLI
+	api.RemoteConnectionsApi
+}
+
+type CreateParams struct {
+	Name             string
+	Description      string
+	OrgID            string
```
danxmoran, line 19: Even though the API requires an org ID, most CLI commands support passing the org name. The command's impl has to look up the org ID as its 1st step in this case.

clients/remote/remote.go
```diff
@@ -0,0 +1,97 @@
...
+	if params.Description != "" {
+		body.Description = &params.Description
+	}
+
+	if body.OrgID == "" && c.ActiveConfig.Org == "" {
+		return clients.ErrMustSpecifyOrg
+	}
+
+	if body.OrgID == "" {
+		body.OrgID = c.ActiveConfig.Org
```
danxmoran, line 46: `c.ActiveConfig.Org` is a name, not an ID. You'll need to use the get-orgs API to look up the name. [Here](https://github.com/influxdata/influx-cli/blob/main/clients/bucket/create.go#L62-L76) is an example.

clients/remote/remote.go
```diff
@@ -0,0 +1,97 @@
...
+		RemoteAPIToken:   params.RemoteAPIToken,
+		RemoteOrgID:      params.RemoteOrgID,
+		AllowInsecureTLS: params.AllowInsecureTLS,
+	}
+
+	if params.Description != "" {
+		body.Description = &params.Description
+	}
+
+	if body.OrgID == "" && c.ActiveConfig.Org == "" {
```
danxmoran, line 41: Recommend moving this check to the very top of the function, so we bail out immediately.

clients/remote/remote.go
```diff
@@ -0,0 +1,97 @@
...
+		var v interface{}
+		if opts.remote != nil {
+			v = opts.remote
+		} else {
+			v = opts.remotes
+		}
+		return c.PrintJSON(v)
+	}
+
+	headers := []string{"Remote ID", "Name"}
```
danxmoran, line 75: I would include:
* `Organization ID`
* `Remote URL`
* `Remote Organization ID`
* `Allow Insecure TLS`

cmd/influx/remote.go
```diff
@@ -22,13 +23,60 @@ func newRemoteCmd() cli.Command {
...
+			},
+			&cli.StringFlag{
+				Name:        "remote-url",
+				Usage:       "The url for the remote database",
+				Required:    true,
+				Destination: &params.RemoteURL,
+			},
+			&cli.StringFlag{
+				Name:        "remote-api-token",
+				Usage:       "The API token for the remote database",
```
danxmoran, line 58: Should be required as well

cmd/influx/remote.go
```diff
@@ -22,13 +23,60 @@ func newRemoteCmd() cli.Command {
...
+			},
+			&cli.StringFlag{
+				Name:        "remote-org-id",
+				Usage:       "The ID of the remote organization",
+				Required:    true,
+				Destination: &params.RemoteOrgID,
+			},
+			&cli.BoolFlag{
+				Name:        "allow-insecure-tls",
+				Usage:       "Allows insecure TLS",
```
danxmoran, line 69: Would add an explicit default of `false`

danxmoran (commented):
There doesn't seem to be much interesting business logic in the core client here, so I don't think heavy unit-testing will be worth the time spent.

cmd/influx/remote.go
mcfarlm3: I think the default is already false, by using `BoolFlag{}` rather than `BoolTFlag{}`? I don't see a value field for `BoolFlag{}`, similar to how you would set a default value for a string flag.

clients/remote/remote.go
```diff
@@ -0,0 +1,115 @@
...
+		var v interface{}
+		if opts.remote != nil {
+			v = opts.remote
+		} else {
+			v = opts.remotes
+		}
+		return c.PrintJSON(v)
+	}
+
+	headers := []string{"Org ID", "Remote URL", "Remote Org ID", "Allow Insecure TLS"}
```
danxmoran, line 91: Oops, miscommunication: I think we want these _and_ the `ID` and `Name` headers you had here previously.

cmd/influx/remote.go
danxmoran: Ah, then this is ok :+1:

danxmoran (approved):

PR #272, opened 2021-09-20
fix: less confusing overloads of 'token' in help

lesam:
Closes https://github.com/influxdata/influxdb/issues/20619


lesam:
After changes:

```
> ./influx v1 auth list
ID                      Description             Username        v2 User Name    v2 User ID              Permissions
082d006fa1324000        my first v1 token       dude            sam             082d001dd1324000        [write:orgs/9aebecf29c3ffd0d/buckets/8bcaafa0c7b887c3]
> ./influx v1 auth delete -h
NAME:
   influx v1 auth delete - Delete authorization

USAGE:
   influx v1 auth delete [command options] [arguments...]

COMMON OPTIONS:
   --host value                     HTTP address of InfluxDB [$INFLUX_HOST]
   --skip-verify                    Skip TLS certificate chain and host name verification [$INFLUX_SKIP_VERIFY]
   --configs-path value             Path to the influx CLI configurations [$INFLUX_CONFIGS_PATH]
   --active-config value, -c value  Config name to use for command [$INFLUX_ACTIVE_CONFIG]
   --http-debug
   --json                           Output data as JSON [$INFLUX_OUTPUT_JSON]
   --hide-headers                   Hide the table headers in output data [$INFLUX_HIDE_HEADERS]
   --token value, -t value          Token to authenticate request [$INFLUX_TOKEN]

OPTIONS:
   --id value        The ID of the authorization
   --username value  The username of the authorization

> ./influx v1 auth delete --username dude
ID                      Description             Username        v2 User Name    v2 User ID              Permissions                   Deleted
082d006fa1324000        my first v1 token       dude            sam             082d001dd1324000        [write:orgs/9aebecf29c3ffd0d/buckets/8bcaafa0c7b887c3] true
```

lesam:
Note `COMMON OPTIONS` which is now across all subcommand help.

DStrand1:
This seems to address #132 as well

clients/apply/apply.go
```diff
@@ -46,23 +46,9 @@ type ResourceFilter struct {
...
-		orgName := params.OrgName
-		if orgName == "" {
-			orgName = c.ActiveConfig.Org
-		}
-		orgs, err := c.GetOrgs(ctx).Org(orgName).Execute()
-		if err != nil {
-			return fmt.Errorf("failed to get ID for org %q: %w", orgName, err)
-		}
-		if len(orgs.GetOrgs()) == 0 {
-			return fmt.Errorf("no orgs found with name %q: %w", orgName, err)
```
lesam, line 63: @davidby-influx pointed out that this is broken since `err == nil`

dgnorton (approved):

clients/cli.go
```diff
@@ -38,3 +42,32 @@ func (c *CLI) PrintTable(headers []string, rows ...map[string]interface{}) error
...
+		return paramOrgId, nil
+	}
+	orgName := paramOrgName
+	if orgName == "" {
+		orgName = c.ActiveConfig.Org
+	}
+	if orgName == "" {
+		return "", ErrMustSpecifyOrg
+	}
+	res, err := orgApi.GetOrgs(ctx).Org(orgName).Execute()
```
DStrand1, line 57: This has a small behavior change now from a few of the commands, like `secret` which. would query `GetOrgs(ctx)` for first the passed parameter, then the active config if none found, but in this case it will only use the active config if no user parameter is specified. I'm not sure which is the "better" option but something to point out I think for discussion

lesam: I much prefer not having to know the state of the InfluxDB instance in order to know how the command will behave, and I don't think the old behaviour is documented, so I strongly feel the updated version is better.

cmd/influx/global.go
```diff
@@ -170,65 +179,115 @@ func getAPINoToken(ctx *cli.Context) *api.APIClient {
...
+		customFunc["iscommon"] = func(flag cli.Flag) bool {
+			if _, ok := flag.(*CommonBoolFlag); ok {
+				return true
+			}
+			if _, ok := flag.(*CommonStringFlag); ok {
+				return true
+			}
+			return false
+		}
+		customFunc["has"] = func(flag cli.Flag) bool {
```
danxmoran, line 281: I don't see this custom func used anywhere, is it needed?

cmd/influx/global.go
```diff
@@ -170,65 +179,115 @@ func getAPINoToken(ctx *cli.Context) *api.APIClient {
...
+		customFunc := make(map[string]interface{})
+		customFunc["iscommon"] = func(flag cli.Flag) bool {
+			if _, ok := flag.(*CommonBoolFlag); ok {
+				return true
+			}
+			if _, ok := flag.(*CommonStringFlag); ok {
+				return true
+			}
+			return false
+		}
```
danxmoran, line 280: Style question: would it be cleaner / more idiomatic to use a type-switch instead of the repeated `if`s?

cmd/influx/global.go
lesam: Nope, missed some cruft. I was going to have an extra function to skip the `COMMON FLAGS` section when none were present, but I don't think that's actually the case for any command.

cmd/influx/global.go
lesam: Yeah that would be nicer, fixed

clients/cli.go
lesam: Thanks for pointing out the behaviour change though!

danxmoran (approved):
:shipit:

PR #278, opened 2021-09-22
feat: generate clients for replications APIs

danxmoran:
Just adding the codegen for now to unblock integration tests on the server-side.


mcfarlm3 (approved):
Looks good!

PR #279, opened 2021-09-22
chore: include enum values in openapi generated code

lesam:
In the API code the enums are still just `string` types, but this additionally exposes the list of possible values in the generated go code.


lesam:
Closing as this needs some more work

PR #281, opened 2021-09-23
fix: allow explicit empty request bodies

danxmoran:
For some use-cases (i.e. sending a no-op write request) it's useful to be able to specify a body of `[]byte{}`. This refactors our codegen template to allow that.


DStrand1 (approved):

mcfarlm3 (approved):
This change seems simple, and makes sense to me. Looks good!

PR #285, opened 2021-09-27
feat: all-access and operator token from CLI

lesam:
Closes https://github.com/influxdata/influxdb/issues/22510


api/extras/README.md
```diff
@@ -0,0 +1,11 @@
+# Influx CLI - extra files supporting HTTP client
```
danxmoran, line 1: We have `api/contract/overrides/` already, do we also need to add `api/extras/`? Could we merge the two and add this note about our changes to enum generation in `api/templates/README.md`?

cmd/influx/auth.go
```diff
@@ -65,86 +70,40 @@ func newCreateCommand() cli.Command {
...
-		&cli.BoolFlag{
-			Name:        "read-checks",
-			Usage:       "Grants the permission to read checks",
-			Destination: &params.ReadCheckPermission,
-		},
-		&cli.BoolFlag{
-			Name:        "write-notificationRules",
-			Usage:       "Grants the permission to create notificationRules",
-			Destination: &params.WriteNotificationRulePermission,
+			Name:        "operator",
```
danxmoran, line 73: Instead of having `--operator` and `--all-access` flags, what do you think of making them be separate subcommands (i.e. `auth create --all-access`)? I ask because it doesn't really make sense to do something like `auth create --all-access --write-buckets abcd`

api/extras/README.md
lesam: I don't think so - we don't want to update the main swagger in such a way that it causes the default generators to actually validate the enum (to avoid breaking cloud/other tooling).

The compromise is to have `api/cli-extras.gen.yml`, which has this (only) to convince the openapi generator that these types are schemas that need full enum generation, rather than in-line enums that only get minimal codegen:

```
components:
  schemas:
    ResourceEnumOSS:
      $ref: "./openapi/src/common/schemas/Resource.yml#/properties/type"
    ResourceEnumCloud:
      $ref: "./openapi/src/cloud/schemas/Resource.yml#/properties/type"
```

Importantly, if there are any references to Resource.yml higher in the file, these would not be fully expanded in `api/cli-extras.gen.yml` and so the enum code is not generated because then they count as inline. It is quite annoying.

cmd/influx/auth.go
lesam: I see them as groups of permissions - in theory we could have another grouping of `--dashboard-read` or something. And composing groups of permissions with extra individual ones makes sense to me.

lesam: And something like `--all-access --read-orgs` might make sense even now.

lesam: @russorat FYI

lesam: In the end I updated this so `--operator` and `--all-access` are still flags, but it now gives an error if you try to use either of these together with finer-grained permission flags.

api/extras/.openapi-generator-ignore
```diff
@@ -0,0 +1 @@
+README.md
```
danxmoran, line 1: Can we add `utils.gen.go` to this list? AFAIK it's a duplicate of the existing `api/utils.gen.go`

api/templates/model_enum.mustache
```diff
@@ -0,0 +1,75 @@
+// {{{classname}}} {{#description}}{{{.}}}{{/description}}{{^description}}the model '{{{classname}}}'{{/description}}
```
danxmoran, line 1: Can you add a line to `api/templates/README.md` describing the change added to this template?

clients/auth/auth.go
```diff
@@ -37,43 +39,71 @@ type printParams struct {
...
 	clients.OrgParams
 	User        string
 	Description string
 
-	WriteUserPermission bool
-	ReadUserPermission  bool
-
-	WriteBucketsPermission bool
-	ReadBucketsPermission  bool
+	ResourcePermissions []*ResourcePermission
```
danxmoran, line 55: Why `[]*ResourcePermission` instead of `[]ResourcePermission`?

clients/auth/auth.go
```diff
@@ -106,86 +143,81 @@ func (c Client) Create(ctx context.Context, params *CreateParams) error {
...
+		}
+		resources, err := c.GetResources(ctx).Execute()
+		if err != nil {
+			return err
+		}
+		for _, r := range resources {
+			if r == string(extras.RESOURCEENUMCLOUD_ORGS) {
+				// orgs are handled specifically - all access is for a single org, not global access to all orgs
+				permissions = append(permissions, api.Permission{
+					Action: ReadAction,
```
danxmoran, line 194: Sanity-checking: should all-access also get `WriteAction` on the org?

cmd/influx/auth.go
```diff
@@ -21,9 +23,33 @@ func newAuthCommand() cli.Command {
 	}
 }
 
+func helpText(perm string) struct{ readHelp, writeHelp string } {
+	var helpOverrides = map[string]struct{ readHelp, writeHelp string }{
+		"user":      {"perform read actions against organization users", "perform mutative actions against organization users"},
+		"buckets":   {"perform mutative actions against organization buckets", "perform mutative actions against organization buckets"},
```
danxmoran, line 29:
```suggestion
		"buckets":   {"perform read actions against organization buckets", "perform mutative actions against organization buckets"},
```

cmd/influx/auth.go
```diff
@@ -21,9 +23,33 @@ func newAuthCommand() cli.Command {
...
+		"telegrafs": {"read telegraf configs", "create telegraf configs"},
+		"orgs":      {"read organizations", "create organizations"},
+		"dbrps":     {"read database retention policy mappings", "create database retention policy mappings"},
```
danxmoran, line 32: Why are these ones `create` instead of `perform mutative actions`?

cmd/influx/auth.go
```diff
@@ -21,9 +23,33 @@ func newAuthCommand() cli.Command {
...
+	help.readHelp = "Grants the permission to " + help.readHelp
+	help.writeHelp = "Grants the permission to " + help.writeHelp
+	return help
+}
+
 func newCreateCommand() cli.Command {
 	var params auth.CreateParams
 	flags := append(commonFlags(), getOrgFlags(&params.OrgParams)...)
+
+	// default: create and update foo / reed foo
```
danxmoran, line 52:
```suggestion
	// default: create and update foo / read foo
```
Is this a misplaced comment? Should it go up where `help.readHelp` and `help.writeHelp` are set?

api/extras/.openapi-generator-ignore
lesam: I think we want to keep it regenerating because the generated code in this repo needs a `utils.gen.go` in its own module (otherwise we'd need to hack the generated code to use the one in the other module)

clients/auth/auth.go
lesam: `append` does copies but I give away pointers to the internal state (for the argument setters).

clients/auth/auth.go
lesam: Specifically no - at least, the ui repo does not do this. See `src/authorizations/utils/permissions.ts` .

```
  if (t === 'orgs') {
    // 'orgs' used to only have read permissions so that's all we'll give again.
    // In production, orgs with an orgID returns a permissions error.
    return [
      {
        action: 'read' as 'read',
        resource: {type: t, id: orgID},
      },
    ]
  }
```

cmd/influx/auth.go
lesam: The default is `create or update` - see line 40 below.

For these I am just preserving the old help text, since I didn't have an opinion on the 'correct' help text.

api/templates/model_enum.mustache
lesam: Done

danxmoran (approved):

PR #286, opened 2021-09-28
feat: all-access and operator token from CLI

lesam:
Closes #22510


lesam:
Opened in error, real one is https://github.com/influxdata/influx-cli/pull/285

PR #289, opened 2021-10-05
feat: added functionality for `replication create` command

### LINKED ISSUE #246: Implement `replication streams create` subcommand
This command should register a new replication stream

mcfarlm3:
Closes #246 

The `replication create` subcommand registers a new replication stream. This configures an existing bucket to replicate all future points to one or more remote InfluxDB instances. 

**Required Flags**

`--name, -n` (Name for new replication stream)
`-- remote-id` (Remote connection the new replication stream should send data to)
`--local-bucket` (ID of local bucket data should be replicated from)
`--remote-bucket` (ID of remote bucket data should be replicated to)

**Optional Flags**

`--description, -d` (Description for new replication stream)
`--org-id` (The ID of the local organization)
`--org, -o` (The name of the local organization)
`--max-queue-bytes` (Max queue size in bytes, default 67108860)

**Note:** A user would need to start with a registered remote connection before creating a replication stream for it. The `influx remote` command can be used to view and manage remote connections.

### DIFF
--- a/clients/replication/replication.go
+++ b/clients/replication/replication.go
@@ -3,6 +3,7 @@ package replication
 import (
 	"context"
 	"fmt"
+
 	"github.com/influxdata/influx-cli/v2/api"
 	"github.com/influxdata/influx-cli/v2/clients"
 )
@@ -14,14 +15,14 @@ type Client struct {
 }
 
 type CreateParams struct {
-	Name             string
-	Description      string
-	OrgID			 string
-	OrgName			 string
-	RemoteID         string
-	LocalBucketID	 string
-	RemoteBucketID   string
-	MaxQueueSize	 int64
+	Name           string
+	Description    string
+	OrgID          string
+	OrgName        string
+	RemoteID       string
+	LocalBucketID  string
+	RemoteBucketID string
+	MaxQueueSize   int64
 }
 
 func (c Client) Create(ctx context.Context, params *CreateParams) error {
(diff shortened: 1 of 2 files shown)


clients/replication/replication.go
```diff
@@ -0,0 +1,104 @@
...
+		var v interface{}
+		if opts.replication != nil {
+			v = opts.replication
+		} else {
+			v = opts.replications
+		}
+		return c.PrintJSON(v)
+	}
+
+	headers := []string{"ID", "Name", "Description", "Org ID", "Remote Connection ID", "Local Bucket", "Remote Bucket", "Max Queue Size Bytes"}
```
danxmoran, line 76:
```suggestion
	headers := []string{"ID", "Name", "Org ID", "Remote ID", "Local Bucket ID", "Remote Bucket ID", "Max Queue  Bytes", "Current Queue Bytes", "Latest Status Code"}
```
* For some reason we don't include "Description" in many other commands, so would leave it out here too
* Adjust some names to match more closely to the API / other CLI terminology
* Add a few read-only properties of the replication that people will eventually want to see

cmd/influx/replication.go
```diff
@@ -22,13 +23,71 @@ func newReplicationCmd() cli.Command {
...
+			},
+			&cli.StringFlag{
+				Name:        "org, o",
+				Usage:       "The name of the local organization",
+				EnvVar:      "INFLUX_ORG",
+				Destination: &params.OrgName,
+			},
+			&cli.StringFlag{
+				Name:        "remote-id",
+				Usage:       "Remote connection ID new replication stream will be registered with",
```
danxmoran, line 58:
```suggestion
				Usage:       "Remote connection the new replication stream should send data to",
```

cmd/influx/replication.go
```diff
@@ -22,13 +23,71 @@ func newReplicationCmd() cli.Command {
...
+				Required:    true,
+				Destination: &params.RemoteID,
+			},
+			&cli.StringFlag{
+				Name:        "local-bucket",
+				Usage:       "Local bucket ID for new replication stream",
```
danxmoran, line 64:
```suggestion
				Usage:       "ID of local bucket data should be replicated from",
```

cmd/influx/replication.go
```diff
@@ -22,13 +23,71 @@ func newReplicationCmd() cli.Command {
...
+				Required:    true,
+				Destination: &params.LocalBucketID,
+			},
+			&cli.StringFlag{
+				Name:        "remote-bucket",
+				Usage:       "Remote bucket ID for new replication stream",
```
danxmoran, line 70:
```suggestion
				Usage:       "ID of remote bucket data should be replicated to",
```

cmd/influx/replication.go
```diff
@@ -22,13 +23,71 @@ func newReplicationCmd() cli.Command {
...
+				Required:    true,
+				Destination: &params.RemoteBucketID,
+			},
+			&cli.Int64Flag{
+				Name:        "max-queue",
+				Usage:       "Max queue size in bytes",
+				Value:       67108860,
```
danxmoran, line 77: Is this value code-gen'd out of the API spec anywhere, for us to reuse here? If not could you add a comment pointing to the part of the spec it's copied from?

PR #294, opened 2021-10-06
fix: properly scope org and user permissions in `auth create`

### LINKED ISSUE #293: `auth create` creates improperly-scoped permissions for `--<action>-orgs` and `--<action>-users`
Passing any of these flags to the `auth create` command creates permissions that don't make sense:
* `--read-orgs`
* `--write-orgs`
* `--read-users`
* `--write-users`

In each of these cases, the CLI creates a permission like:
```
{
  "type": <orgs|users>,
  "orgId": <ID of org passed to `auth create`, or org from CLI config>
}
```

Permissions with that structure don't make sense because orgs and users are both top-level resources, not nested under any other org. We should special-case the logic for those flags so that they leave out the `orgId` field from the generated permissions.

danxmoran:
Closes #293


clients/auth/auth.go
```diff
@@ -152,10 +152,21 @@ func (c Client) Create(ctx context.Context, params *CreateParams) error {
 			actions = append(actions, WriteAction)
 		}
 
+		permOrgID := orgID
+		// Orgs and users are both "top-level" resources, not nested under any other org.
+		// Read/write permissions on those resource types must also be top-level to have an effect.
+		switch resourcePermission.Name {
+		case string(extras.RESOURCEENUMCLOUD_ORGS):
+			permOrgID = ""
+		case string(extras.RESOURCEENUMCLOUD_USERS):
```
lesam, line 161: If we're dealing with the users permission properly, should we also deal with non-operator users that want permission to edit themselves but not other users?

lesam: (see special handling for the all-access permissions)

danxmoran: I think I'd prefer to punt those situations to a follow-up PR and generalize the current special case of `--<action>-bucket <id1>,<id2>,...` to a set of `--<action>-<type> <id1>,<id2>,...` flags

PR #295, opened 2021-10-06
chore: expose tabwriter so InfluxDB can use it

lesam:


tabwriter/tabwriter_test.go
```diff
@@ -4,7 +4,7 @@ import (
 	"bytes"
 	"testing"
 
-	"github.com/influxdata/influx-cli/v2/internal/tabwriter"
+	"github.com/influxdata/influx-cli/v2/tabwriter"
```
danxmoran, line 7: Can we stick it under `pkg/`? That's where all the other random utils currently live

lesam: Good point. Done

danxmoran (approved):

PR #296, opened 2021-10-06
feat: added functionality for replication list command

### LINKED ISSUE #245: Implement `replication-streams list` subcommand
This command should list all existing replication streams and relevant information for each

mcfarlm3:
Closes #245 

This command lists all currently registered replication streams associated with the local org. It also includes options to filter the output by name, remote ID and/or local bucket ID. The output fields include ID, Name, Org ID, Remote ID, Local Bucket ID, Remote Bucket ID, Current Queue Bytes, Max Queue Bytes, and Latest Status Code for each replication stream.

See options below:

Screen Shot 2021-10-06 at 1 02 38 PM

### DIFF
--- a/clients/replication/replication.go
+++ b/clients/replication/replication.go
@@ -2,6 +2,7 @@ package replication
 
 import (
 	"context"
+	"errors"
 	"fmt"
 
 	"github.com/influxdata/influx-cli/v2/api"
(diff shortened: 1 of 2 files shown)


clients/replication/replication.go
```diff
@@ -56,6 +57,55 @@ func (c Client) Create(ctx context.Context, params *CreateParams) error {
...
+
+	orgID, err := c.GetOrgId(ctx, params.OrgID, params.OrgName, c.OrganizationsApi)
+	if err != nil {
+		return err
+	}
+
+	// set up params
+	req := c.GetReplications(ctx)
+
+	req = req.OrgID(orgID)
```
danxmoran, line 78: Nitpick: you could chain this `.OrgID` after the `.GetReplications` call above

danxmoran (approved):
Nice!

PR #298, opened 2021-10-07
feat: added functionality for replication delete command

### LINKED ISSUE #248: Implement `replication-streams delete` subcommand
This command should delete an existing replication stream, given its ID. This command will also delete data still in the stream, if any.

mcfarlm3:
Closes #248 

This command deletes an existing replication stream, given its ID. This command will also delete data still in the stream, if any. `id` is the only option for this command, and it is required.

Screen Shot 2021-10-07 at 10 04 36 AM

**Note:** This PR also changes the naming of the `remote-id` flag for `remote update` and `remote delete` subcommands.
`--remote-id, -id` --> `--id, -i`
This was done to maintain consistency with other `influx` CLI commands.

### DIFF
--- a/clients/replication/replication.go
+++ b/clients/replication/replication.go
@@ -105,16 +105,14 @@ func (c Client) List(ctx context.Context, params *ListParams) error {
 }
 
 func (c Client) Delete(ctx context.Context, replicationID string) error {
+	// get replication stream via ID
 	connection, err := c.GetReplicationByID(ctx, replicationID).Execute()
 	if err != nil {
-		return fmt.Errorf("failed to delete replication stream %q: %w", replicationID, err)
+		return fmt.Errorf("could not find replication stream with ID %q: %w", replicationID, err)
 	}
 
-	req := c.DeleteReplicationByID(ctx, replicationID)
-
 	// send delete request
-	err = req.Execute()
-	if err != nil {
+	if err := c.DeleteReplicationByID(ctx, replicationID).Execute(); err != nil {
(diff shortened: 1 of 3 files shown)


clients/replication/replication.go
```diff
@@ -104,6 +104,29 @@ func (c Client) List(ctx context.Context, params *ListParams) error {
 	return c.printReplication(printOpts)
 }
 
+func (c Client) Delete(ctx context.Context, replicationID string) error {
+	connection, err := c.GetReplicationByID(ctx, replicationID).Execute()
+	if err != nil {
+		return fmt.Errorf("failed to delete replication stream %q: %w", replicationID, err)
```
danxmoran, line 110:
```suggestion
		return fmt.Errorf("could not find replication stream with ID %q: %w", replicationID, err)
```

clients/replication/replication.go
```diff
@@ -104,6 +104,29 @@ func (c Client) List(ctx context.Context, params *ListParams) error {
...
+	}
+
+	req := c.DeleteReplicationByID(ctx, replicationID)
+
+	// send delete request
+	err = req.Execute()
+	if err != nil {
```
danxmoran, line 117: Nitpick: could collapse from line 113 to here into a single line, nested inside the conditional.:
```go
if err := c.DeleteReplicationByID(ctx, replicationID).Execute(); err != nil {
    //stuff
}
```

danxmoran (approved):

PR #305, opened 2021-10-13
fix: embed tzdata into Windows builds

### LINKED ISSUE #304: Migrate cross-test to use cross-builder executor
1. Add a new CI job to cross-build test binaries, persisting the outputs in the workflow's workspace
2. Update existing native-test jobs to run the test binaries instead of building from scratch

Benefits of doing this:
1. Keeps this repo in-sync with the way things are done in `influxdb`
2. Reduces the need for us to deal with strangeness in Circle's Windows and Mac executors

danxmoran:
Closes #304

This PR was originally meant to be build-only, aligning a piece of the `influx-cli` CI setup with `influxdb`. In the process I rediscovered golang/go#38453. We work around that issue here in the same way we did in influxdata/influxdb#22447, setting a build tag when `GOOS=windows` so TZ data is embedded in the binary.

### DIFF
--- a/.circleci/config.yml
+++ b/.circleci/config.yml
@@ -1,25 +1,29 @@
 version: "2.1"
 
 executors:
+  cross-builder:
+    docker:
+      # NOTE: To upgrade the Go version, first push the upgrade to the cross-builder Dockerfile in the edge repo,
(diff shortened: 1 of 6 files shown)


davidby-influx (approved):
LGTM, but I recommend @codyshepherd also review

PR #308, opened 2021-10-18
build: upgrade protobuf library

DStrand1:
Upgrades Influx CLI to use the new `google.golang` protobuf library over the abandoned `gogo` protobuf library


clients/restore/restore.go
```diff
@@ -370,7 +370,7 @@ func (c Client) restoreOrg(ctx context.Context, name string) (string, error) {
 	return *orgs.GetOrgs()[0].Id, nil
 }
 
-func bucketToDBI(b br.ManifestBucketEntry) br.DatabaseInfo {
+func bucketToDBI(b br.ManifestBucketEntry) *br.DatabaseInfo {
```
danxmoran, line 373: Why do these now need to return pointers?

internal/backup_restore/bolt.go
```diff
@@ -62,7 +62,7 @@ func ExtractBucketMetadata(boltPath string) ([]api.BucketMetadataManifest, error
 	// Read raw metadata needed to construct a manifest.
 	var buckets []influxdbBucketSchema
 	orgNamesById := map[string]string{}
-	dbInfoByBucketId := map[string]DatabaseInfo{}
+	dbInfoByBucketId := map[string]*DatabaseInfo{}
```
danxmoran, line 65: Same here, why pointers now?

clients/restore/restore.go
DStrand1: To both: with the new protobuf library, leaving those as before would cause issues with `go vet`, stating `return copies lock value` as `proto.MessageState` contains a mutex now

PR #315, opened 2021-10-27
revert: "feat: return error if API token required but not found"

slai:
Reverts influxdata/influx-cli#310

I've discovered an issue with this PR, and I can't think of a way to fix it without breaking changes, so I'm proposing to revert for now.

---

The `influx telegrafs` command is a command with sub-commands. Running it as `influx telegrafs` tells the CLI to list telegraf configurations, and running `influx telegrafs create` tells the CLI to run the `create` sub-command instead.

This creates a problem when specifying the token using the `--token` or `-t` arg because args are scoped to the nearest preceding command, and both commands have `withApi(true)` specified as `Before` functions.

Therefore running -

```
❯ ./influx telegrafs -t test create
Error: API token required. Pass `--token`, set `INFLUX_TOKEN`, or use `influx config set --token` to specify your token value
```

causes the `withApi(true)` before function specified on the `create` subcommand to fail because it has no `--token` arg value (`-t test` was scoped to the `telegrafs` command).

And running -

```
❯ ./influx telegrafs create -t test
Error: API token required. Pass `--token`, set `INFLUX_TOKEN`, or use `influx config set --token` to specify your token value
```

causes the `withApi(true)` before function specified on the `telegrafs` command to fail. Even `./influx telegrafs create -h` fails.

This previously worked because there was no arg validation when creating the API client, so when running `./influx telegrafs create -t test` -

* the `withApi(true)` before function on the `telegrafs` command produced an invalid API client with no token set, 
* **but** the subsequent `withApi(true)` before function on the `create` command _overrode_ that with a valid API client that had the token set. 

To prove this, specifying the token twice works -

```
❯ ./influx telegrafs -t test create -t test
Error: must specify org ID or org name
```

As does using INFLUX_TOKEN as environment variables are global. But neither of these are good solutions though.

Some potential solutions -

* rewrite commands so there are no commands that also have sub-commands, e.g. move the `telegrafs` command to `telegrafs list`
* somehow make the `withApi` before function only create an API client if there are no more commands to be parsed

The first one is a breaking change, and I haven't quite worked out how to do the second option, so to ensure a broken release doesn't happen, I propose we revert for now.


danxmoran (approved):
Unfortunate, but makes sense to revert for now. I'll make a follow-up issue to investigate restoring the nicer error message

PR #312, opened 2021-10-21
feat: nicer error when command not found

slai:
Currently, when you try to use an unrecognized command, you get -

```
❯ ./influx hello
No help topic for 'hello'
```

This isn't particularly clear, and you might wonder how you triggered the help mode.

This PR changes that to something more explicit -

```
❯ ./influx hello
Error: command "hello" not recognized. Run `./influx --help` to see the list of commands
```

PR #316, opened 2021-11-02
fix: return "unknown command" instead of "no help topic" error when unknown (sub)command is passed

danxmoran:
Follow-up to #312

### DIFF
--- a/cmd/influx/bucket_schema.go
+++ b/cmd/influx/bucket_schema.go
@@ -36,8 +36,9 @@ func getBucketSchemaClient(ctx *cli.Context) bucket_schema.Client {
 
 func newBucketSchemaCmd() cli.Command {
 	return cli.Command{
-		Name:  "bucket-schema",
-		Usage: "Bucket schema management commands",
+		Name:   "bucket-schema",
+		Usage:  "Bucket schema management commands",
+		Before: middleware.NoArgs,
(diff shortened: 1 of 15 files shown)


cmd/influx/bucket_schema.go
```diff
@@ -36,8 +36,9 @@ func getBucketSchemaClient(ctx *cli.Context) bucket_schema.Client {
 
 func newBucketSchemaCmd() cli.Command {
 	return cli.Command{
-		Name:  "bucket-schema",
-		Usage: "Bucket schema management commands",
+		Name:   "bucket-schema",
+		Usage:  "Bucket schema management commands",
+		Before: middleware.NoArgs,
```
lesam, line 41: NoArgs is a weird name for this - I would call it something like `SubcommandValid`

lesam (approved):
One nitpick about existing naming

cmd/influx/bucket_schema.go
danxmoran: The name & error message mirror the `cobra` equivalent. It's meant to say "this command doesn't take any positional args, so if you see one try parsing it as a subcommand" which then leads to "unknown command" errors... agreed it's not always the most ideal way to think about it.

slai:
@danxmoran LGTM, thanks for taking this over and merging it

PR #325, opened 2021-11-10
feat: add flags for remotes, replications, and functions to `auth create`

### LINKED ISSUE #314: Add read & write flags for remotes & replications to `influx auth create`

danxmoran:
Closes #314

The flags are currently hidden while we finish the server-side work.


cmd/influx/auth.go
```diff
@@ -46,6 +46,16 @@ func helpText(perm string) struct{ readHelp, writeHelp string } {
 	return help
 }
 
+func hidden(perm string) bool {
+	var hiddenTypes = map[string]struct{}{
+		"functions":    {},
```
danxmoran, line 51: Hiding functions here since we haven't implemented the related CLI commands, but maybe it's ok to leave visible?

lesam: But I could see this argued either way, so I don't feel strongly.

api/contract/cli.yml
```diff
@@ -12,7 +12,7 @@ paths:
   /setup:
     $ref: "./openapi/src/common/paths/setup.yml"
   /write:
-    $ref: "./openapi/src/common/paths/write.yml"
```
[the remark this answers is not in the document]
danxmoran: It can. `common/paths/write.yml` got moved to `oss/paths/write.yml`, and `cloud/paths/write.yml` was created to handle different sets of potential error codes between the two. We eventually need to reconcile the two but I didn't want to do that work as part of this.

cmd/influx/auth.go
danxmoran: We've been proactively hiding the replication commands so I think it makes sense to follow suit here (and piggy-back on the existing un-hide ticket). For functions I could see making it visible now since it's live in Cloud.

lesam (approved):

PR #330, opened 2021-11-16
feat: add `drop-non-retryable-data` to replications commands

wbaker85:
Part of https://github.com/influxdata/influxdb/issues/22880

Also see: https://github.com/influxdata/influxdb/pull/22885

### DIFF
--- a/clients/replication/replication.go
+++ b/clients/replication/replication.go
@@ -4,7 +4,6 @@ import (
 	"context"
 	"errors"
 	"fmt"
-	"strings"
 
 	"github.com/influxdata/influx-cli/v2/api"
 	"github.com/influxdata/influx-cli/v2/clients"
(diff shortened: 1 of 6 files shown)


clients/replication/replication.go
```diff
@@ -46,6 +60,15 @@ func (c Client) Create(ctx context.Context, params *CreateParams) error {
 		body.Description = &params.Description
 	}
 
+	if params.DropNonRetryableData != "" {
+		setB, err := stringToBool(params.DropNonRetryableData)
+		if err != nil {
+			return err
+		}
+
+		body.SetDropNonRetryableData(setB)
```
wbaker85, line 69: Using this setter keeps us from needing to create a pointer to a boolean

clients/replication/replication.go
```diff
@@ -17,12 +18,25 @@ type Client struct {
...
-	LocalBucketID  string
-	RemoteBucketID string
-	MaxQueueSize   int64
+	Name                 string
+	Description          string
+	RemoteID             string
+	LocalBucketID        string
+	RemoteBucketID       string
+	MaxQueueSize         int64
+	DropNonRetryableData string
```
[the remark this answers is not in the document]
danxmoran: I would push back on this, I think it's a more idiomatic CLI experience to have both `--flag` and `--no-flag` for bool toggles. Those could be consolidated down to a single `*bool` in this struct without a ton of custom code at the `cmd/` layer.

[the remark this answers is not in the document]
wbaker85: Made updates so that this works with a `--drop-non-retryable-data` and `--no-drop-non-retryable-data` flag now. I didn't end up using a `*bool` here since I would need to store the value for both flags and compute the correct pointer value for those somewhere, and it seemed a little easier to do that in the client code.

clients/replication/replication.go
```diff
@@ -17,12 +17,14 @@ type Client struct {
...
-	RemoteBucketID string
-	MaxQueueSize   int64
+	Name                   string
+	Description            string
+	RemoteID               string
+	LocalBucketID          string
+	RemoteBucketID         string
+	MaxQueueSize           int64
+	DropNonRetryableData   bool
+	NoDropNonRetryableData bool
```
danxmoran, line 27: You could probably get away with only 1 flag here (for whatever isn't the default in the API), but not a blocker if you want to keep 2 flags for symmetry with `update`.

clients/replication/replication.go
```diff
@@ -212,3 +232,23 @@ func (c Client) printReplication(opts printReplicationOpts) error {
...
+	}
+
+	if noDropNonRetryableData {
+		return boolPtr(false), nil
+	}
+
+	return nil, nil
+}
+
+func boolPtr(b bool) *bool {
```
danxmoran, line 252: I think there's already a function for this auto-generated under `api`.

danxmoran (approved):

danxmoran (approved):

PR #336, opened 2021-11-22
feat: add ExecuteWithHttpInfo methods for generated API

### LINKED ISSUE #335: Provide a way to get the full `http.Response` from API requests
The API template used by this project provides a simplified way of handling API responses by omitting the full `http.Reponse`: https://github.com/influxdata/influx-cli/blob/main/api/templates/README.md

There are, however, situations where it is useful to receive the `http.Response` as part of an API request, such as dealing with `retry-after` headers on responses from the cloud `write` API. There should be a way to make an API request and get the `http.Response`, perhaps as part of a `ExecuteWithHttpInfo` method, which would be consistent with how some other client generators work.

For reference, the original template for the go client generator is here, which has the `Execute` methods return the `http.Response` in addition to the resource object and an error: https://github.com/OpenAPITools/openapi-generator/blob/master/modules/openapi-generator/src/main/resources/go/api.mustache

wbaker85:
Closes #335

This creates additional methods for generated API clients that can be used to obtain the full HTTP response object. This is needed for things like the OSS 2.x replications feature, which must inspect response codes and headers to determine appropriate retry actions. No existing CLI operations should be impacted by these additional methods, although they could use them in the future if needed.

### DIFF
--- a/api/templates/README.md
+++ b/api/templates/README.md
@@ -19,7 +19,11 @@ multiple locations.
 * Add checks for `isResponseBinary` to directly return the raw `*http.Response`, instead of
   pulling the entire body into memory and transforming it into an `*os.File`
 * GUnzip non-binary response bodies before unmarshalling when `Content-Encoding: gzip` is set
-* Remove `*http.Response`s from the return values of generated `Execute` operations & provide a separate method `ExecuteWithHttpInfo` for returning the `*http.Response` if needed
+* Remove `*http.Response`s from the return values of generated `Execute` operations & provide a separate method
+  `ExecuteWithHttpInfo` for returning the `*http.Response` if needed. The standard `Execute` operation should normally be
+  used unless specific header values or response codes need to be examined with the `ExecuteWithHttpInfo`. The response
+  body cannot be read from the `*http.Reponse` returned from ExecuteWithHttpInfo` as it will have already have been read
+  and closed; access to the response body content should be achieved through the returned response model if applicable
 
 `client.mustache`
 * Removed use of `golang.org/x/oauth2` to avoid its heavy dependencies
--- a/api/templates/README.md
+++ b/api/templates/README.md
@@ -22,7 +22,7 @@ multiple locations.
 * Remove `*http.Response`s from the return values of generated `Execute` operations & provide a separate method
   `ExecuteWithHttpInfo` for returning the `*http.Response` if needed. The standard `Execute` operation should normally be
   used unless specific header values or response codes need to be examined with the `ExecuteWithHttpInfo`. The response
-  body cannot be read from the `*http.Reponse` returned from ExecuteWithHttpInfo` as it will have already have been read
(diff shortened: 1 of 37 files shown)


api/templates/README.md
```diff
@@ -19,7 +19,7 @@ multiple locations.
 * Add checks for `isResponseBinary` to directly return the raw `*http.Response`, instead of
   pulling the entire body into memory and transforming it into an `*os.File`
 * GUnzip non-binary response bodies before unmarshalling when `Content-Encoding: gzip` is set
-* Remove `*http.Response`s from the return values of generated operations
+* Remove `*http.Response`s from the return values of generated `Execute` operations & provide a separate method `ExecuteWithHttpInfo` for returning the `*http.Response` if needed
```
lesam, line 22: Why do we want this? Sounds like a way to get around the openapi spec?

lesam: Response codes should already be available in the returned errors if openapi is set up correctly. There's also OpenAPI spec for "header parameters", I'd prefer we use it rather than blanket returning all headers: https://swagger.io/docs/specification/describing-parameters/

wbaker85: Not trying to get around the openapi spec, which does define header parameters appropriately for the `/write` endpoint, for ex: https://github.com/influxdata/openapi/blob/7aac839df611dc786182b38e81cf239a3e47184c/contracts/cloud.yml#L6122-L6129

The reasons for wanting to do this are related to the OpenAPI code generation:
1. We need to be able to access headers like the one above for the replications remote write call if an "over quota" error like that is received. The only way to get response headers with the code-gen'd API clients is through the returned `http.Reponse` object from the from the `Execute` operations.
2. The codegen treats any 2XX response as a success and doesn't return an error. For the `/write` endpoint, only a `204` is a success. A `200` response is an error, and we can't treat it like a success to prevent dataloss. A proper InfluxDB endpoint wouldn't return a non-204 response on success, but I know the first thing that I did to test this was to point it at a requestbin and was surprised that it seemed to be validating/writing successfully to something returning a 200 response code, so it feels like we need to be more strict here.

The default API codegen templates actually include the `http.Response` object as part of the response. If we had not modified them specifically to _not_ include that, this PR wouldn't have been necessary. But it was a nice simplification for the CLI code to not have to discard the unused `http.Response` since accessing any specific header values hasn't (yet) been necessary for the CLI.

As an alternative...for (1) we could just use the same exp. backoff for 429 errors as everything else - this probably wouldn't be too big of a problem. (2) does seem like a deficiency in the codegen, but fixing that appears to be non-trivial since that behavior is the standard from the code generator we are using.

lesam: re 1: The openapi schema generator supports a section for headers, I'd prefer we insert generated code to pass through the headers we support rather than passing through the Response object. We can look at this together if you like, I can show you how to look at the parameters that are passed to the template generator.

re 2:  `A 200 response is an error` sounds like a bug. Does both cloud and OSS do this? Why?

lesam (dismissed):

lesam (approved):

PR #338, opened 2021-12-03
fix: Use influx-debug-id header

mkmik:
Influxdata cloud2 uses a different mechanism to enable tracing.
You need to pass a `influx-debug-id` header (the value is ignored) and if the response contains
`trace-sampled: true` then the response's `trace-id` header contains a valid trace ID that can be
used to jump to the trace in jaeger.

If the request has been sampled, we print the trace-id to stderr.

Since modern terminals allow to click on hyperlinks, we could avoid a copy&paste of the trace id
if the `influx` binary wrote a the link to the jeager trace.
However, we don't want to hardcode the URL of the internal jaeger endpoint in the binary.

I tried adding a wrapper script on my machine that would just capture the trace-id and render it, but it's
hard to do without messing with the ordering of the stderr and stdout lines.
Thus I preferred adding a secret env var that controls the formatting of the trace-id and can be used
by our devs to generate links to our internal jaeger instance.

```console
$ export INFLUX_CLI_TRACE_PRINT_PREFIX="Trace: https://jaeger.my.cloud/trace/"
$ influx query --trace-debug-id 123 -f query.flux
Trace: https://jaeger.my.cloud/trace/f39a69354a3acca6
Result: _result
Table: keys: [_measurement]
...
```

### DIFF
--- a/api/client.gen.go
+++ b/api/client.gen.go
@@ -245,6 +245,13 @@ func (c *APIClient) callAPI(request *http.Request) (*http.Response, error) {
 		}
 		log.Printf("\n%s\n", string(dump))
 	}
+	if resp.Header.Get("trace-sampled") == "true" {
+		tracePrefix := "trace-id: "
+		if prefix, found := os.LookupEnv("INFLUX_CLI_TRACE_PRINT_PREFIX"); found {
+			tracePrefix = prefix
+		}
+		fmt.Fprintf(os.Stderr, "%s%s\n", tracePrefix, resp.Header.Get("trace-id"))
+	}
 	return resp, err
 }
 
--- a/api/configuration_default.go
+++ b/api/configuration_default.go
@@ -37,6 +37,7 @@ func NewAPIConfig(params ConfigParams) *Configuration {
 		// use codegen'd logic to set the header on every HTTP request. Early versions of the CLI
 		// used that technique, and we found it to be error-prone and easy to forget during testing.
 		apiConfig.DefaultHeader["Zap-Trace-Span"] = *params.TraceId
+		apiConfig.DefaultHeader["influx-debug-id"] = *params.TraceId
 	}
 	apiConfig.Debug = params.Debug
 
--- a/api/templates/client.mustache
+++ b/api/templates/client.mustache
@@ -191,6 +191,13 @@ func (c *APIClient) callAPI(request *http.Request) (*http.Response, error) {
 		}
 		log.Printf("\n%s\n", string(dump))
 	}
+	if resp.Header.Get("trace-sampled") == "true" {
+		tracePrefix := "trace-id: "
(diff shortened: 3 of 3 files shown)


primeroz:
I understand this is a public project so i think 

* We should make this a flag 
* We should keep the default to the common `jaeger-debug-id`

mkmik:
> I understand this is a public project so i think
> 
> * We should make this a flag
> * We should keep the default to the common `jaeger-debug-id`

The flag already exists and it's called "--trace-debug-id"

This PR does two things:

1. Makes that flag emit both the header honoured by OSS influxdb and the header honoured by cloud2

2. Adds this extra env var which is used to report the trace id returned by cloud2

Would it make it less confusing if I split this in two PRs?

primeroz (approved):

primeroz:
> Would it make it less confusing if I split this in two PRs?


nah, i obviously did not understand the change :)

api/client.gen.go
```diff
@@ -245,6 +245,13 @@ func (c *APIClient) callAPI(request *http.Request) (*http.Response, error) {
 		}
 		log.Printf("\n%s\n", string(dump))
 	}
+	if resp.Header.Get("trace-sampled") == "true" {
```
wbaker85, line 248: This file is code-gen'd from a template, so you're going to need to edit the template and then generate this file.

It should be pretty easy to add this to https://github.com/influxdata/influx-cli/blob/4c0fae3a4c0d77df0fcbd91583df8b246d488e30/api/templates/client.mustache#L173

Then  you can run `make openapi` to re-generate this client file.

mkmik:
```console
$ ./etc/generate-openapi.sh
Error opening file "/api/contract/openapi/src/oss/paths/health.yml"
ENOENT: no such file or directory, open '/api/contract/openapi/src/oss/paths/health.yml'
```

mkmik: ah gitsubmodules...

mkmik:
```console
$ cat .gitmodules
[submodule "openapi"]
        path = api/contract/openapi
        url = ../openapi
```

wbaker85: https://github.com/influxdata/influx-cli/blob/main/api/contract/README.md#updating-the-api-contract

mkmik: yeah I didn't expect it to be a relative path to another checkout. I'll send another PR

mkmik: UH, TIL git knows how to handle relative paths on remotes too now? wow

mkmik: anyway sorry for the noise; now the change is in the moustache template and I issued `make openapi`

wbaker85 (approved):

PR #345, opened 2021-12-21
chore: update openapi ref to latest

wbaker85:
Updates the repo to use the latest spec from https://github.com/influxdata/openapi

Generally straightforward, but required some small changes to the telegrafs client.


mcfarlm3 (approved):
This looks good to me!

PR #347, opened 2021-12-23
feat: enhanced error messages for cloud and oss specific commands

### LINKED ISSUE #142: Improve error messages even more for OSS- or cloud-only commands

wbaker85:
Closes #142

This will use the actual `X-Influxdb-Build` header returned by the server to determine the cause for an error.

The mechanism for flagging commands as cloud/OSS only is also improved. Middleware is used to add information to the `cli.Context` for host-specific commands, and an exit error handler middleware determines if a host-specific command was used, and if the wrong host type is present in the response header. This should make it much easier to keep track of which commands are specific to which host.

### DIFF
--- a/api/api_authorizations.gen.go
+++ b/api/api_authorizations.gen.go
@@ -131,30 +131,10 @@ type AuthorizationsApi interface {
 	 * @return Authorization
 	 */
 	PostAuthorizationsExecuteWithHttpInfo(r ApiPostAuthorizationsRequest) (Authorization, *_nethttp.Response, error)
-
-	// Sets additional descriptive text in the error message if any request in
-	// this API fails, indicating that it is intended to be used only on OSS
-	// servers.
-	OnlyOSS() AuthorizationsApi
-
-	// Sets additional descriptive text in the error message if any request in
-	// this API fails, indicating that it is intended to be used only on cloud
-	// servers.
-	OnlyCloud() AuthorizationsApi
 }
 
 // AuthorizationsApiService AuthorizationsApi service
 type AuthorizationsApiService service
-
-func (a *AuthorizationsApiService) OnlyOSS() AuthorizationsApi {
-	a.isOnlyOSS = true
-	return a
-}
-
-func (a *AuthorizationsApiService) OnlyCloud() AuthorizationsApi {
-	a.isOnlyCloud = true
-	return a
-}
 
 type ApiDeleteAuthorizationsIDRequest struct {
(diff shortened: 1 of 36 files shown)


wbaker85:
Example error messages

#### Cloud-only commands run on OSS:

New:
```
influx bucket-schema list --bucket-id 1234123412341234
Error: InfluxDB Cloud-only command used with InfluxDB OSS host
```

Old:
```
influx bucket-schema list --bucket-id 1234123412341234
Error: failed to find bucket "1234123412341234": InfluxDB Cloud-only command failed: 404 Not Found: bucket not found
```

#### OSS-only commands run on Cloud:
New:
```
influx ping       
Error: InfluxDB OSS-only command used with InfluxDB Cloud host
```

Old:
```
influx ping
```

New:
```
influx remote list
Error: InfluxDB OSS-only command used with InfluxDB Cloud host
```

Old:
(`remote` was not listed as an OSS-only command, although it should have been)
```
influx remote list             
Error: failed to get remote connections: 404 Not Found: path not found
```

mcfarlm3 (approved):
Took me a bit of time to trace my way through this, but it looks good to me! A lot cleaner than before, and this will be a good pattern to have for adding any future OSS-only or Cloud-only commands.

wbaker85:
Renamed the test `TestMain_HostSpecificErrors` to `TestApp_HostSpecificErrors` to avoid confusion with the actual [`TestMain` function](https://pkg.go.dev/testing#hdr-Main) prior to merging.

PR #348, opened 2021-12-23
fix: duration parser shows duration missing units on error

### LINKED ISSUE #171: Duration parser error when no units are provided is not clear
When a duration integer is provided but no units are provided with it, the resulting error message from the duration parser is not clear, and almost implies that there is some kind of missing additional information:

```
influx bucket update -i aef36e903cc987a1 -r 1209600000000000
Error: duration is missing a unit:
```

A better error message might include the list of acceptable units, similar to how it would look if an invalid duration unit is provided.

wbaker85:
Closes #171

Before: 
```
influx bucket update -i 853de12dd738bd7d -r 11234 
Error: duration is missing a unit: 
```

After:
```
influx bucket update -i 853de12dd738bd7d -r 11234            
Error: duration is missing a unit: 11234
```

### DIFF
--- a/pkg/duration/parser.go
+++ b/pkg/duration/parser.go
@@ -93,7 +93,8 @@ func parseSignedDuration(text string) (durations, error) {
 }
 
 // parseDuration will convert a string into components of the duration.
-func parseDuration(lit string) (durations, error) {
+func parseDuration(input string) (durations, error) {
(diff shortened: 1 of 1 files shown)


mcfarlm3 (approved):
LGTM!

PR #349, opened 2021-12-23
build: automatically generate changelog

### LINKED ISSUE #324: Automate CHANGELOG generation
Port the auto-changelog tooling from `influxdb`

wbaker85:
Closes #324

Other than the `changelog` job, this required providing specific names for the test binaries generated for the various platforms. Otherwise the workspace could not be mounted in the release job due to multiple jobs having concurrently added files with the same name to the workspace.


.gitignore
```diff
@@ -11,6 +11,10 @@ dist/
 # Test binary, built with `go test -c`
 *.test
 
+# Directories mounted during the release process in CI
```
wbaker85, line 14: Otherwise gorelease errors out due to a dirty git state

.goreleaser.yml
```diff
@@ -1,8 +1,10 @@
 project_name: influx-cli
 
-# Do not make github release
+# Do not make github release or generate a changelog
 release:
   disable: true
+changelog:
```
wbaker85, line 6: Not 100% necessary, but might be helpful to prevent confusion due to the generate goreleaser CHANGELOG, which is not relevant since we are generating our own.

DStrand1 (approved):
LGTM!

PR #355, opened 2022-01-24
fix: jsonnet files in `influx apply` return a 422 error

### LINKED ISSUE #351: `apply` with `-e jsonnet` fails
Applying a `jsonnet` template causes an error when using the CLI with a cloud host, and will soon cause an error with an OSS host: https://github.com/influxdata/influxdb/pull/23030

To reproduce, the file at https://github.com/influxdata/influxdb/blob/master/pkger/testdata/bucket_associates_labels.jsonnet can be used with  `influx apply -e jsonnet -f bucket_associates_labels.jsonnet` with the file saved locally, or even `influx apply -e jsonnet -u https://github.com/influxdata/influxdb/blob/master/pkger/testdata/bucket_associates_labels.jsonnet` for a hosted file.

Since server-side jsonnet parsing is now disabled, we will need to parse it in the client to be able to handle jsonnet templates.

DStrand1:
Closes #351 

This isn't the "ideal" fix, as much of the code in `client.gen.go`/`client.mustache` is unnecessary and overcomplicated, but this PR fixes the `jsonnet` file ext being passed along to the server, which is simply incorrect since we are correctly sending JSON-serialized data. The `Content-Type` header is properly being set as `application/json`, so there should be no risk of data confusion.

I will open a follow-up issue to fix the panic on receiving a 422 error, and to refactor the code further to not send the data as a form and follow a better API call structure

### DIFF
--- a/pkg/template/source.go
+++ b/pkg/template/source.go
@@ -87,7 +87,7 @@ func SourcesFromPath(path string, recur bool, encoding Encoding) ([]Source, erro
 		}
 
 		sources[i] = Source{
-			Name:     path,
+			Name:     strings.TrimSuffix(path, filepath.Ext(path)),
 			Encoding: encoding,
 			Open: func(context.Context) (io.ReadCloser, error) {
 				return os.Open(path)


jdstrand (commented):
I can confirm that this fixes the issue. The server doesn't need the filename as it will look at the Content-Type (honestly, it is bizarre to me that we are sending the local file to the server via a multipart form; `influx` should just be hitting `/api/v2/templates/apply` via POST with `application/json` and be done with it, but that can happen in a followup).

Some things I tested (note, `~/code/influx` is my locally built `influx` with this PR applied):
```
# All of this worked:
$ ~/code/influx apply -f ./ok.yaml
$ ~/code/influx apply -f ./ok.json
$ ~/code/influx apply -f ./ok.jsonnet
$ ~/code/influx apply -f ./ok.json -e jsonnet  # note, json is a subset of jsonnet
$ cat ./ok.yaml | ~/code/influx apply  # defaults to yaml
$ cat ./ok.yaml | ~/code/influx apply -e yaml
$ cat ./ok.json | ~/code/influx apply -e json
$ cat ./ok.jsonnet | ~/code/influx apply -e jsonnet

# These correctly failed:
$ ~/code/influx apply -f ./ok.yaml -e json
$ ~/code/influx apply -f ./ok.json -e yaml
$ ~/code/influx apply -f ./ok.jsonnet -e yaml
$ cat ./ok.json | ~/code/influx apply  # default of yaml is wrong for json
$ cat ./ok.json | ~/code/influx apply -e yaml
$ cat ./ok.yaml | ~/code/influx apply -e json
$ cat ./ok.yaml | ~/code/influx apply -e jsonnet
$ cat ./ok.jsonnet | ~/code/influx apply -e json  # jsonnet is a superset of json
$ cat ./ok.jsonnet | ~/code/influx apply -e yaml
```

It would be great if some table driven tests could be constructed to do the above so we can rely on manual testing less.

I'm submitting this as a 'Comment' instead of 'Request changes' so the PR isn't blocked on me. If the tests are added, I'd be happy to approve.

jdstrand:
> I'm submitting this as a 'Comment' instead of 'Request changes' so the PR isn't blocked on me. If the tests are added, I'd be happy to approve.

I'd like to see the git commit message to have some supporting comments. Eg, `pkg/template/source.go:SourcesFromPath()` is modified (fine), but `cmd/influx/apply.go:newApplyCmd()` and `cmd/influx/template.go:parseSources()` both use `pkg/template/source.go:SourcesFromPath()`. It isn't immediately obvious that the change in this PR is correct for both so adding some justification would help the reviewer. Thanks!

pkg/template/source.go
```diff
@@ -87,7 +87,7 @@ func SourcesFromPath(path string, recur bool, encoding Encoding) ([]Source, erro
 		}
 
 		sources[i] = Source{
-			Name:     path,
+			Name:     strings.TrimSuffix(path, filepath.Ext(path)),
```
lesam, line 90: This still doesn't work for something horrendous like `foo.jsonnet.jsonnet`. I have another idea, going to close this and open a new PR.

lesam:
going to close this and open a new one

PR #356, opened 2022-01-27
fix: 422 error handling panic

DStrand1:
Previously, the error handling code in `client.gen.go` would panic when a 422 status code was received, because it prioritized calling `e.model.Error()`, but the 422 error only set the error in `e.error`, leading to the panic. This PR reverses the checked order for the error, though required a new panic case for if both `e.error` and `e.model.Error()` are errorless.

### DIFF
--- a/api/templates/client.mustache
+++ b/api/templates/client.mustache
@@ -580,10 +580,13 @@ func (e GenericOpenAPIError) BuildHeader() string {
 
 // Error returns non-empty string if there was an error.
 func (e GenericOpenAPIError) Error() string {
+	if e.error != "" {
+		return e.error
+	}
 	if e.model != nil {
 		return e.model.Error()
 	}
-	return e.error
+	panic("errorless GenericOpenAPIError used as an error!")
 }
 
 // Model returns the unpacked model of the error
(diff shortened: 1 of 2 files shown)


api/templates/client.mustache
```diff
@@ -580,10 +580,13 @@ func (e GenericOpenAPIError) BuildHeader() string {
...
 // Error returns non-empty string if there was an error.
 func (e GenericOpenAPIError) Error() string {
+	if e.error != "" {
+		return e.error
+	}
 	if e.model != nil {
 		return e.model.Error()
 	}
-	return e.error
+	panic("errorless GenericOpenAPIError used as an error!")
```
mcfarlm3, line 589: I think this change should fix the panic you were seeing with the 422 response, but I am wondering in what situation would we be calling the `e.Error()` function, but both `e.error` and `e.model.Error()` are errorless? We would expect to always have an error in one or the other, right?

lesam (commented):
Awaiting discussion on https://github.com/influxdata/idpe/issues/12979

DStrand1:
Superseded by #362

Issue #233, opened 2021-08-20
influx apply cannot reapply the same template with the same stack id


russorat:
i was trying to apply an update to an existing stack but i get the error 
`Error: template has conflicts with existing resources and cannot safely apply`

Since i am applying the exact same template, i would expect it to apply but not make any changes. this seems to be a regression from the previous cli.

```
(base) ➜  ~ influx apply -c docker --skip-verify -u https://raw.githubusercontent.com/influxdata/community-templates/master/sample-data/sample-data.yml --stack-id 08050271a709f000 --force true
BUCKETS    +add | -remove | unchanged
+-----+--------------------+------------------+---------------+------------------+-------------+-------------+------------------+
| +/- |   METADATA NAME    |        ID        | RESOURCE NAME | RETENTION PERIOD | DESCRIPTION | SCHEMA TYPE | NUM MEASUREMENTS |
+-----+--------------------+------------------+---------------+------------------+-------------+-------------+------------------+
|     | sample-data-bucket | ddddb22d7f3b2e60 | sample_data   | 168h0m0s         |             | implicit    |                0 |
+-----+--------------------+------------------+---------------+------------------+-------------+-------------+------------------+
|                                                                                                   TOTAL    |        1         |
+-----+--------------------+------------------+---------------+------------------+-------------+-------------+------------------+

TASKS    +add | -remove | unchanged
+-----+-------------------+------------------+-------------------+-------------+-------+
| +/- |   METADATA NAME   |        ID        |   RESOURCE NAME   | DESCRIPTION | CYCLE |
+-----+-------------------+------------------+-------------------+-------------+-------+
|     | fetch-sample-data | 08050271ad5d1000 | Fetch Sample Data |             |       |
+-----+-------------------+------------------+-------------------+-------------+-------+
|                                                                     TOTAL    |   1   |
+-----+-------------------+------------------+-------------------+-------------+-------+

Error: template has conflicts with existing resources and cannot safely apply
(base) ➜  ~ influx apply -c docker --skip-verify -u https://raw.githubusercontent.com/influxdata/community-templates/master/sample-data/sample-data.yml --force true
BUCKETS    +add | -remove | unchanged
+-----+--------------------+----+---------------+------------------+-------------+-------------+------------------+
| +/- |   METADATA NAME    | ID | RESOURCE NAME | RETENTION PERIOD | DESCRIPTION | SCHEMA TYPE | NUM MEASUREMENTS |
[...]

PR #358, opened 2022-02-04
fix: template apply uses better diff checking

lesam:
fix: when statestatus is 'remove', mark as removed

fix: template apply uses better diff checking

Closes: https://github.com/influxdata/influx-cli/issues/233

Previously, we did a DeepEqual of all the returned data about each changed entity,
but due to our template overrides that is not actually all the information
available for each entity. So we marked trivial things as 'conflicts' (e.g.
telegraf config ID's 'changing' from the empty string to the real, current value)
while not catching important conflicts like flux script changes in checks and tasks.

Changes to make things more straightforward:

 * Change the --force behaviour to be more similar to `apt install`, where even
    in non-interactive mode `--force yes` is required to bypass the prompt to apply.

 * Before, there were two stages of diff checking - once to print diffs, and once
    after the 'Yes/No' prompt. If any conflicts were detected in the second check,
    the user got an inscrutable error message that did not highlight what the difference
    was or how to force it to apply. `--force conflict` was required to avoid this error.
    Instead, we now have simplified to `--force yes` to bypass the 'Yes/No' prompt, and
    we never do a second stage of diff checking.

 * Because we do not currently properly account for more complicated diffs (e.g.
    flux tasks), we now assume in the diff printing that every object has changes,
    except for Labels, Buckets, Variables, and Label Mappings. This could be improved
    in the future.

### DIFF
--- a/clients/apply/apply.go
+++ b/clients/apply/apply.go
@@ -4,7 +4,6 @@ import (
 	"context"
 	"errors"
 	"fmt"
-	"reflect"
 	"strconv"
 	"time"
 
@@ -33,7 +32,6 @@ type Params struct {
 	Filters []ResourceFilter
 
 	Force              bool
-	OverwriteConflicts bool
 	Quiet              bool
 	RenderTableColors  bool
 	RenderTableBorders bool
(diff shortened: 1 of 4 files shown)


DStrand1 (approved):
LGTM!

> Because we do not currently properly account for more complicated diffs (e.g.
flux tasks), we now assume in the diff printing that every object has changes,
except for Labels, Buckets, Variables, and Label Mappings. This could be improved
in the future.

Would you mind opening an issue for this?

lesam:
Issue opened ☝️

PR #362, opened 2022-02-16
chore: update to API spec for new template endpoint response

DStrand1:
Fixes the panic in #351, but not the root cause of the issue, which is fixed by #355

Updates the CLI to the latest OpenAPI version, which importantly contains a change for the `api/v2/template` endpoint. The `422` response syntax has changed slightly, and as a result fixes the crash in #351.


api/contract/overrides/schemas/TemplateSummaryError.yml
```diff
@@ -1,16 +1,21 @@
...
+  diff:
+    $ref: "./TemplateSummaryDiff.yml"
+  errors:
     type: array
     items:
-      type: integer
-    x-go-field-type: '[]*int'
-required: [kind, reason, fields, indexes]
+      $ref: "./TemplateSummaryErrors.yml"
+required: [message, code]
```
DStrand1, line 21: Ignore the messy diff here.. the CLI had a file named `TemplateSummaryError` already that was renamed to `TemplateSummaryErrors`, as OpenAPI now has a `TemplateSummaryError` file

api/api_templates.gen.go
```diff
@@ -193,7 +193,7 @@ func (a *TemplatesApiService) ApplyTemplateExecuteWithHttpInfo(r ApiApplyTemplat
 		newErr.body = localVarBody
 		newErr.error = localVarHTTPResponse.Status
 		if localVarHTTPResponse.StatusCode == 422 {
-			var v TemplateSummary
+			var v TemplateSummaryError
```
lesam, line 196: Does TemplateSummaryError implement SetMessage 'properly'?

api/contract/overrides/schemas/TemplateSummaryError.yml
lesam: Do we still need all the overrides for templates now that this (hopefully) matches the non-override API spec more closely?

clients/template/template.go
```diff
@@ -81,7 +81,7 @@ func (c Client) Validate(ctx context.Context, params *ValidateParams) error {
 	}
 
 	if apiErr, ok := err.(api.GenericOpenAPIError); ok {
-		if summary, ok := apiErr.Model().(*api.TemplateSummary); ok {
+		if summary, ok := apiErr.Model().(*api.TemplateSummaryError); ok {
```
lesam, line 84: Any way that this could do less type assertion? For example, does apiErr.Error() now return something more reasonable, or could we make that happen?

api/api_templates.gen.go
lesam: (i.e. if we get an error back from `a.client.decode`, will that be reflected properly in the output?

lesam (approved):
I have some comments but none of them should block merging.

api/api_templates.gen.go
DStrand1: Yes, it implements it via code generation now

clients/template/template.go
DStrand1: With type assertion to `TemplateSummaryError`:
```
Error: template failed validation:
        kind=unknown field= reason="invalid kind"
```
With type assertion to `GenericOpenAPIError`:
```
Error: template failed validation:
        kind=unknown field= reason="invalid kind"
```
Without type assertion:
```
Error: failed to validate template: kind=unknown field= reason="invalid kind"
```

Seems like we can get away with just asserting to `GenericOpenAPIError` now

api/error.go
```diff
@@ -121,18 +123,6 @@ func (o *TemplateSummary) Error() string {
 	return strings.Join(errMsg, "\n\t")
```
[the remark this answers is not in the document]
DStrand1: I'll open a tech-debt issue about cleaning this up in the future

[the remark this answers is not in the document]
DStrand1: Issue opened here: https://github.com/influxdata/influx-cli/issues/363

api/contract/overrides/schemas/TemplateSummaryError.yml
DStrand1: `influx apply` still heavily depends on these overrides for generating `diff` blocks, as well as some other places for printing out the `resources` block. I hope this can one day be cleaned up, but its outside of the scope of this PR I think

PR #372, opened 2022-04-11
feat: Add --extra-http-header flag

mkmik:
Some influxdb feature flags can be controlled by headers; let's add a way to play with those headers without forcing us back into raw `curl`.

### DIFF
--- a/cmd/influx/global.go
+++ b/cmd/influx/global.go
@@ -17,15 +17,16 @@ import (
 )
 
 const (
-	tokenFlagName       = "token"
-	hostFlagName        = "host"
-	skipVerifyFlagName  = "skip-verify"
-	traceIdFlagName     = "trace-debug-id"
-	configPathFlagName  = "configs-path"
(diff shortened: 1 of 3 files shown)


cmd/influx/global.go
```diff
@@ -212,6 +229,11 @@ func coreFlags() []cli.Flag {
 			Hidden: true,
 			EnvVar: "INFLUX_TRACE_DEBUG_ID",
 		}},
+		&CommonStringSliceFlag{cli.StringSliceFlag{
```
lesam, line 232: Would be nice to have usage, don't know if it needs to be hidden - I like the feature.

samhld:
@lesam @mkmik This will only affect Cloud users, right? If so, this should be documented in the CLI docs and the cloud docs...but _not_ the OSS docs, right?

mkmik:
It's a hidden option FWIW. Mostly useful for internal uses

PR #373, opened 2022-04-13
build: upgrade to Go 1.18.1

DStrand1:
Upgrades to Go 1.18.1

- `// +build ...` directives are no longer needed, only `//go:build` is used now
- `reflect.Ptr` was renamed to `reflect.Pointer` (`reflect.PtrTo()` was similarly renamed `reflect.PointerTo()`, though we had no usages of this function)
    - The old name still exists, but will be deprecated in a future release (and is non-breaking to change now)
- A few sections of code were moved to `strings.Cut()` for slight code cleanliness


davidby-influx (approved):
LGTM

Issue #371, opened 2022-03-24
environment variable $INFLUX_ORG_ID inappropriately overrides command line option


rogpeppe:
As a general convention, command-line options should override environment variables because
environment variables are non-specific, whereas command-line options apply only to that single
command invocation.

This does not appear to be the case with the `influx` command.

Here's an example:

```
$ export INFLUX_ORG_ID=844910ece80be8bc
$ influx write -b devbucket -o 9c5955fc99a60b8f --host https://rog.remocal.influxdev.co/ < /tmp/x.lp
Error: failed to write data: 404 Not Found: organization not found
$ unset INFLUX_ORG_ID
$ influx write -b devbucket -o 9c5955fc99a60b8f --host https://rog.remocal.influxdev.co/ < /tmp/x.lp
$
```

PR #377, opened 2022-05-13
fix: add mutual exclusion for OrgId and OrgName params

candrewlee14:
### Before
Environment variables did get correctly overridden by the command line flags already, but the problem was when a user would pass the `-o` flag (or `INFLUX_ORG` env var) and the `INFLUX_ORG_ID` env var (or `--org-id` flag), it wasn't made clear that only the OrgId was being used. OrgName would be ignored.

### After
Now, any command that uses the org flags enforces only one can be used, not both.
When both flags are passed together, the user will see an error like this: 
```
Error: ambiguous org: use OrgId or OrgName, but not both. OrgID: 844910ece80be8bc, OrgName: 9c5955fc99a60b8f
```

Closing #371


davidby-influx (approved):
LGTM

PR #379, opened 2022-05-23
feat: Add --version option to influx query

jsternberg:
The --version option will execute a query against influxdb that returns
the current flux version on the server.

It uses the query from the docs:
https://docs.influxdata.com/influxdb/v2.2/query-data/flux/flux-version/

### DIFF
--- a/clients/helpers.go
+++ b/clients/helpers.go
@@ -51,3 +51,8 @@ func ReadQuery(filepath string, args []string) (string, error) {
 		return arg, nil
 	}
 }
+
+const VersionQuery = `import "array"
+import "runtime"
+array.from(rows: [{version: runtime.version()}])
+`
--- a/clients/query/version_printer.go
+++ b/clients/query/version_printer.go
@@ -1,0 +1,34 @@
+package query
+
+import (
+	"fmt"
+	"io"
+
+	"github.com/influxdata/influx-cli/v2/pkg/fluxcsv"
+)
+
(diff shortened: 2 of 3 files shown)


samhld:
Thanks, @jsternberg !  Whoever reviews, can you verify that this flag only applies for the `query` subcommand? I'm fine with having this be universal to `influx` but would then likely need its name changed to something like `--flux-version`.

DStrand1:
> ... can you verify that this flag only applies for the `query` subcommand? I'm fine with having this be universal to `influx` but would then likely need its name changed to something like `--flux-version`.

It is currently only on the `influx query` subcommand, though I do wonder if a new subcommand for this, like `influx flux-version` would be even better. I'll leave it up to you @jsternberg

DStrand1 (approved):
LGTM, assuming we are sticking with `influx query --version` and not some other method of checking the version as mentioned above

samhld:
@DStrand1 I had the same thought. In my experience, a flag usually means I'm modifying/configuring a command in some way. It may make more sense to make this its own subcommand, similar to how `version` is for the version of the DB and the CLI. That said, I'd rather leave that up to be democratically decided by developers with more of a frame reference for standards.

@jsternberg thoughts on making this its own subcommand?

samhld:
@jsternberg would it be trouble to make this its own subcommand instead of a flag to the `query` subcommand?

Issue #331, opened 2021-11-18
Add back the InfluxQL REPL


lesam:
In 1.x, there was an influxql REPL where we could issue influxql commands repeatedly against the same server, without re-typing 'influx' or having annoying shell quoting. We should port the 1.x REPL to the 2.x CLI.

The REPL should give clear error messages when attempting to use 1.x-only InfluxQL features, see https://docs.influxdata.com/influxdb/v2.1/query-data/influxql/?t=InfluxDB+API#influxql-support

lesam:
See also https://github.com/influxdata/influx-cli/issues/331 if we can't get to this in 2.3

PR #386, opened 2022-05-25
feat: add back the InfluxQL REPL

candrewlee14:
Closing #331 
Auxiliary PRs & Issues:
- #393 
- #392 
- #390 
- #389


api/api_authorizations.gen.go
```diff
@@ -208,7 +208,7 @@ func (a *AuthorizationsApiService) DeleteAuthorizationsIDExecuteWithHttpInfo(r A
 		return nil, GenericOpenAPIError{error: err.Error()}
 	}
 
-	localVarPath := localBasePath + "/authorizations/{authID}"
+	localVarPath := localBasePath + "/api/v2/authorizations/{authID}"
```
lesam, line 211: These changes are suprising - why?

api/contract/cli.yml
```diff
@@ -3,123 +3,119 @@ info:
   title: Subset of Influx API covered by Influx CLI
   version: 2.0.0
 servers:
-  - url: /api/v2
```
lesam, line 6: I think this change needs to be made in the upstream openapi repo first - we want to minimize divergence.

lesam: Should we maybe just have a separate API spec for the legacy API instead of trying to put them together?

lesam: I'd be interested in thoughts from @jstirnaman as well

api/api_authorizations.gen.go
candrewlee14: The problem was that the `/query` endpoint in `cli.yml` was used at the root but with the root server as `/api/v2`, but the legacy InfluxQL endpoint is just `/query`. I couldn't have two `/query` endpoints so I needed to move all the `/api/v2` endpoints so that the `/query` becomes `/api/v2/query` and the legacy `/query` can be distinct.
Maybe it wasn't necessary this way, I'm new to using openapi schemas.

api/contract/cli.yml
```diff
@@ -229,6 +225,10 @@ components:
       $ref: "./openapi/src/cloud/schemas/MeasurementSchemaUpdateRequest.yml"
     Query:
       $ref: "./overrides/schemas/Query.yml"
+    InfluxqlCsvResponse:
+      $ref: "./overrides/schemas/InfluxqlCsvResponse.yml"
```
lesam, line 229: This should be in the openapi repo rather than `overrides` I think - we want to shrink the size of `overrides` over time rather than grow it.

lesam: I think we should have a sync conversation then with the three of us - I don't want to pull you in multiple directions.

clients/v1_repl/v1_repl.go
```diff
@@ -0,0 +1,921 @@
...
+func DefaultPersistentQueryParams() PersistentQueryParams {
+	return PersistentQueryParams{
+		Format:    CsvFormat,
+		Precision: "ns",
+	}
+}
+
+func (c *Client) readHistory() []string {
+	// Only load/write history if HOME environment variable is set.
+	var historyDir string
```
lesam, line 88: This repeated code should be extracted - I think we should calculate the history directory at startup similar to the old CLI code

lesam, line 88: This repeated code should be extracted - I think we should calculate the history directory at startup similar to the old CLI code

clients/v1_repl/v1_repl.go
```diff
@@ -0,0 +1,921 @@
...
+	defer c.writeCommandToHistory(cmd)
+	cmdArgs := strings.Split(cmd, " ")
+	switch strings.ToLower(cmdArgs[0]) {
+	case "quit", "exit":
+		color.HiBlack("Goodbye!")
+		os.Exit(0)
+	case "gopher":
+		c.gopher()
+
+	// TODO: Allow user to change API in REPL session
```
lesam, line 327: Is there an issue for this TODO? We should link one here.

lesam, line 327: Is there an issue for this TODO? We should link one here.

clients/v1_repl/v1_repl.go
```diff
@@ -0,0 +1,921 @@
...
+	{Text: "settings", Description: "Output the current shell settings"},
+	{Text: "clear", Description: "Clears settings such as database"},
+	{Text: "exit", Description: "Exit the InfluxQL shell"},
+	{Text: "quit", Description: "Exit the InfluxQL shell"},
+	{Text: "gopher", Description: "Display the Go Gopher"},
+	{Text: "help", Description: "Display help options"},
+	{Text: "format", Description: "Specify the data display format"},
+}
+
+func (c *Client) gopher() {
```
lesam, line 258: lol

clients/v1_repl/v1_repl.go
```diff
@@ -0,0 +1,921 @@
...
+	)
+}
+
+func (c *Client) help() {
+	fmt.Println(`Usage:
+        ` +
+		// ! UNSUPPORTED connect <host:port>   connects to another node specified by host:port
+		// ! UNSUPPORTED auth                  prompts for username and password
+		`pretty                toggles pretty print for the json format
+        use <db_name>         sets current database
```
lesam, line 562: What about chunked responses?

clients/v1_repl/v1_repl.go
```diff
@@ -0,0 +1,921 @@
...
+        format <format>       specifies the format of the server responses: json, csv, or column
+        precision <format>    specifies the format of the timestamp: h, m, s, ms, u or ns
```
lesam, line 564: I think we treat `consistency` similar to `node` - sorry this is a 1.x option not supported on the 2.x CLI.

lesam: Though eventually it would be nice to get backward-compatible support in the CLI for 1.x.

clients/v1_repl/v1_repl.go
```diff
@@ -0,0 +1,921 @@
...
+
+        show databases        show database names
+        show series           show series information
+        show measurements     show measurement information
+        show tag keys         show tag key information
+        show field keys       show field key information
+
+        A full list of influxql commands can be found at:
+        https://docs.influxdata.com/influxdb/latest/query_language/spec/`)
+	// ? Maybe add keybindings information like <CTRL+L> clearing the screen, <CTRL+D> exitting, etc
```
lesam, line 578: Maybe add `insert` as well.

clients/v1_repl/v1_repl.go
```diff
@@ -0,0 +1,921 @@
...
+	for _, db := range dbs {
+		if parsedDb == db {
+			exists := false
+			prevDb := c.Database
+			c.Database = parsedDb
+			rps, _ := c.getRetentionPolicies(context.Background())
+			// discover if the parsedRp is a valid retention policy
+			for _, rp := range rps {
+				if parsedRp == rp || parsedRp == "" {
+					if parsedRp == "" {
```
lesam, line 723: We can hoist the `parsedRp == ""` check out of the loop

clients/v1_repl/v1_repl.go
candrewlee14: I wasn't aware of chunked responses available in the 2.x API, but I meant to leave a TODO in there to look into it further.

clients/v1_repl/v1_repl.go
```diff
@@ -0,0 +1,921 @@
...
+		if err := json.Indent(&buf, []byte(jsonBody), "", "  "); err != nil {
+			color.Red("Unable to prettify json response.")
+			fmt.Println(jsonBody)
+		} else {
+			fmt.Println(buf.String())
+		}
+	}
+}
+
+func (c *Client) outputTable(jsonBody string) {
```
lesam, line 676: What happens if I redirect this to a file? Does the TUI do 'the right thing'?

clients/v1_repl/v1_repl.go
lesam: I really hope we support chunked responses in 2.x - it's important to be able to return a response that doesn't fit it memory (both on the server and client side) by streaming the information one chunk at a time.

lesam (commented):
Great first pass!

clients/v1_repl/v1_repl.go
lesam: It can be a TODO in this PR though.

clients/v1_repl/v1_repl.go
```diff
@@ -0,0 +1,921 @@
...
+		CsvFormat:   c.outputCsv,
+		JsonFormat:  c.outputJson,
+		TableFormat: c.outputTable,
+	}
+	displayFunc := displayMap[c.Format]
+	displayFunc(response)
+}
+
+// This function generates the prompt autocompletions
+func (c *Client) completer(d prompt.Document) []prompt.Suggest {
```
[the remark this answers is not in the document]
candrewlee14: Nope, that's a new feature 😄 Thank you!
Yeah, definitely a big PR, sorry about that...
The main prompt loop runs through the go-prompt library, so do you want me to pull that all out for this PR and make a new PR with that added back in?

clients/v1_repl/v1_repl.go
candrewlee14: Now linked here: https://github.com/influxdata/influx-cli/issues/389

clients/v1_repl/v1_repl.go
candrewlee14: Moved it to a switch statement which `continue`s in the`default` case. Much cleaner imo.

clients/v1_repl/v1_repl.go
```diff
@@ -0,0 +1,912 @@
...
+		return
+	}
+	bucketID, err := c.getBucketIdForDBRP(db, rp)
+	if err != nil {
+		color.Red("Unable to match DBRP to BucketID: %v", err)
+		return
+	}
+	writeReq := c.PostWrite(context.Background()).
+		Bucket(bucketID).
+		Precision(api.WritePrecision(c.Precision)).
```
candrewlee14, line 434: @lesam Looking at this api.WritePrecision, it only expects from `{"ms", "s", "us", "ns"}`.
Should this be using a new`c` field, like `c.WritePrecision` to write, since we want to support `rfc3339` precisions on queries?

api/contract/cli.yml
```diff
@@ -3,123 +3,119 @@ info:
...
+  /private/legacy/authorizations:
+    $ref: "./openapi/src/legacy/paths/legacy_authorizations.yml"
+  /private/legacy/authorizations/{authID}:
+    $ref: "./openapi/src/legacy/paths/legacy_authorizations_authID.yml"
+  /private/legacy/authorizations/{authID}/password:
+    $ref: "./openapi/src/legacy/paths/legacy_authorizations_authID_password.yml"
+  /api/v2/setup:
     $ref: "./openapi/src/common/paths/setup.yml"
-  /write:
+  /api/v2/write:
```
DStrand1, line 22: Are changes like this still needed post-OpenAPI update?

clients/v1_repl/v1_repl.go
```diff
@@ -0,0 +1,912 @@
...
+        https://docs.influxdata.com/influxdb/latest/query_language/spec/`)
+	// ? Maybe add keybindings information like <CTRL+L> clearing the screen, <CTRL+D> exitting, etc
+}
+
+func (c *Client) settings() {
+	w := new(tabwriter.Writer)
+	w.Init(os.Stdout, 0, 1, 1, ' ', 0)
+	fmt.Fprintln(w, "Setting\tValue")
+	fmt.Fprintln(w, "--------\t--------")
+	// TODO: No supported way to access or change host & port in a REPL session
```
DStrand1, line 576: Is this something that might get fixed in a future PR? If so lets make an issue, if not lets take this commented code out (should probably remove the the commented out code either way)

clients/v1_repl/v1_repl.go
```diff
@@ -0,0 +1,912 @@
...
+
+func (c *Client) setPrecision(args []string) {
+	// args[0] is "precision"
+	if len(args) != 2 {
+		color.Red("Expected a precision [ns, u, ms, s, m, or h]")
+		return
+	}
+	precision := args[1]
+	switch precision {
+	case "ns", "u", "µ", "ms", "s", "m", "h":
```
DStrand1, line 640: Are we allowing hour/minute precision? I thought I saw above that we were not but that could have been an unrelated section of code

clients/v1_repl/v1_repl.go
```diff
@@ -0,0 +1,912 @@
...
+type FormatType string
+
+var (
+	CsvFormat   FormatType = "csv"
+	JsonFormat  FormatType = "json"
+	TableFormat FormatType = "table"
+)
+
+func (c *Client) runAndShowQuery(query string) {
+	// ? Should this function have support to guide users trying to use deprecated InfluxQL queries
```
DStrand1, line 492: I would say yes, could even link to the docs page showing which keywords are not allowed if that seems reasonable, that way its a one-time-mistake instead of one time per disallowed keyword

clients/v1_repl/v1_repl.go
```diff
@@ -0,0 +1,912 @@
...
+	dbrpsReq := c.GetDBRPs(context.Background())
+	if c.OrgID != "" {
+		dbrpsReq = dbrpsReq.OrgID(c.OrgID)
+	}
+	if c.OrgName != "" {
+		dbrpsReq = dbrpsReq.Org(c.OrgName)
+	}
+	if c.OrgID == "" && c.OrgName == "" {
+		dbrpsReq = dbrpsReq.Org(c.ActiveConfig.Org)
+	}
```
DStrand1, line 465: This code looks familiar... any way we could reuse the code in `clients/params.go#GetOrgID()` maybe with a new interface instead of directly taking `api.OrganizationsApi` (if it seems like a simple enough change to make)

clients/v1_repl/v1_repl.go
```diff
@@ -0,0 +1,912 @@
...
+	{Text: "USER"},
+	{Text: "USERS"},
+	{Text: "VALUES"},
+	{Text: "WHERE"},
+	{Text: "WITH"},
+	{Text: "WRITE"},
+}
+
+var replKeywords []prompt.Suggest = []prompt.Suggest{
+	// TODO: provide a way to connect to a new endpoint during the same REPL session
```
DStrand1, line 232: Is this covered in #389? If not we should make an issue for it as well

DStrand1 (commented):
Looking great so far! I agree with @lesam about the code-completion portion being in a separate PR for easier reviewing and tracking changes, if it is not too heavily coupled and easy to separate out.

Mostly just a bunch of questions for now

clients/v1_repl/v1_repl.go
candrewlee14: Same issue :)

clients/v1_repl/v1_repl.go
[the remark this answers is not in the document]
candrewlee14: Here's the PR that I separated that out into: https://github.com/influxdata/influx-cli/pull/392
It's set to merge into this branch, hopefully that makes sense to do.

clients/v1_repl/v1_repl.go
candrewlee14: It allowed you to do that, then I think the query engine would give you an error.
I've added the option for `rfc3339` precision, and when you run an `INSERT` command, this error comes back:
`ERR: 400 Bad Request: invalid precision; valid precision units are ns, us, ms, and s`. Is this acceptable or should I do checking/clarification beforehand?

clients/v1_repl/v1_repl.go
candrewlee14: Where did you see that we weren't allowing h/m precision?

clients/v1_repl/v1_repl.go
lesam: I think 'the right thing' is to do what e.g. `less` does with colors, and detect non-tty mode and switch to a simpler output with no colors. Maybe we can dependency-inject colors to be a no-op with no tty?

lesam:
I just tried:

```
./influx v1 repl
InfluxQL Shell dev
Connected to InfluxDB OSS 2.2.0
InfluxQL Shell
> ls
Query failed.
401 Unauthorized: Unauthorized
> ^C^C^Cquit
```

It would be nice if it detected unauthorized errors up front and quit out automatically with a message about setting the token.

lesam:
Also `influx` used to drop direct into the repl with no arguments (in 1.x), so it might be worth displaying some hint about `influx v1 repl` if someone runs a bare `influx`.

lesam:
@samhld it would also be cool if we auto-mapped buckets to v1 databases instead of having users set that up manually - thoughts on that as a follow-up? Otherwise you get, by default:

```
./influx v1 repl
InfluxQL Shell dev
Connected to InfluxDB OSS 2.2.0
InfluxQL Shell
> show databases
name,tags,name
```

lesam:
@candrewlee14 +1 for making the table view the default.

lesam:
@candrewlee14 this looks bad:

```
> show databases
name,tags,name
databases,,monitoring
databases,,telegraf

> use telegraf
Unable to check databases: expected a single array in values array
```

clients/v1_repl/v1_repl.go
```diff
@@ -0,0 +1,723 @@
...
+	}
+	displayFunc := displayMap[c.Format]
+	displayFunc(response)
+}
+
+func (c *Client) help() {
+	fmt.Println(`Usage:
+		pretty                toggles pretty print for the json format
+        use <db_name>         sets current database
+        format <format>       specifies the format of the server responses: json, csv, or column
```
lesam, line 358: Should be `table` not `column`

candrewlee14:
> @candrewlee14 this looks bad:
> 
> 
> 
> ```
> 
> > show databases
> 
> name,tags,name
> 
> databases,,monitoring
> 
> databases,,telegraf
> 
> 
> 
> > use telegraf
> 
> Unable to check databases: expected a single array in values array
> 
> ```

Oh interesting. I didnt expect there to be multiple arrays in values, I thought there was a single array that just had the database names. What does the json look like?

clients/v1_repl/v1_repl.go
candrewlee14: Oh, actually, I think column is the 1.x version of table. I'll implement that. That could also be a good fallback format for non-tty, what do you think?

candrewlee14:
> Is history expected to work?

When I took out go-prompt for autocompletion, it had the support for arrow keys for history, keybindings, etc.

I can re-implement that here, but it might just be worth seeing that in the autocompletion PR.

samhld:
@candrewlee14 on the note of the table, I think we do have to go back to copy the original table view as it handled many columns more readably. That's a case I overlooked and one that won't be that uncommon.

lesam:
> When I took out go-prompt for autocompletion, it had the support for arrow keys for history, keybindings, etc.

That's perfect, just wanted to make sure it was planned in the upcoming PR. Happy to keep this one smaller.

clients/v1_repl/v1_repl.go
lesam: Hmm - looks like it 'just worked' in 1.x, and rfc3339 on query meant `ns` on write:

```
> precision rfc3339
> select * from cpu limit 10
name: cpu
time                 cpu       host              usage_guest usage_guest_nice usage_idle        usage_iowait usage_irq usage_nice usage_softirq usage_steal usage_system        usage_user
----                 ---       ----              ----------- ---------------- ----------        ------------ --------- ---------- ------------- ----------- ------------        ----------
2022-04-28T20:30:20Z cpu-total MacBook-Pro.local 0           0                96.36120847989224 0            0         0          0             0           1.49390752796169    2.1448839926075656
2022-04-28T20:30:20Z cpu0      MacBook-Pro.local 0           0                84.85456369129106 0            0         0          0             0           6.218655967875164   8.926780340833778
2022-04-28T20:30:20Z cpu1      MacBook-Pro.local 0           0                99.8997995991049  0            0         0          0             0           0                   0.10020040080397738
2022-04-28T20:30:20Z cpu10     MacBook-Pro.local 0           0                97.79779779802409 0            0         0          0             0           0.9009009009241942  1.3013013013430397
2022-04-28T20:30:20Z cpu11     MacBook-Pro.local 0           0                99.79999999981374 0            0         0          0             0           0.10000000000218279 0.09999999999763531
2022-04-28T20:30:20Z cpu2      MacBook-Pro.local 0           0                89.6896896894004  0            0         0          0             0           4.504504504479571   5.805805805828698
2022-04-28T20:30:20Z cpu3      MacBook-Pro.local 0           0                99.89989989980658 0            0         0          0             0           0                   0.10010010009782637
2022-04-28T20:30:20Z cpu4      MacBook-Pro.local 0           0                92.29229229209865 0            0         0          0             0           3.1031031030826894  4.604604604600158
2022-04-28T20:30:20Z cpu5      MacBook-Pro.local 0           0                99.89979959910723 0            0         0          0             0           0.10020040079936143 0
2022-04-28T20:30:20Z cpu6      MacBook-Pro.local 0           0                95.29058116037271 0            0         0          0             0           1.8036072143930624  2.905811623119968
> insert m0,tag1=foo bval=asdf 1640775019000000000
ERR: {"error":"unable to parse 'm0,tag1=foo bval=asdf 1640775019000000000': invalid boolean"}

> insert m0,tag1=foo bval=1.2 1640775019000000000
> select * from m0
name: m0
time                 bval tag1
----                 ---- ----
2021-12-29T10:50:19Z 1.2  foo
```

Having it as an error for now seems fine - @samhld do you want a follow-up for something like a 'write precision' separate from read precision?

clients/v1_repl/v1_repl.go
lesam:
```
ERR: 400 Bad Request: invalid precision; valid precision units are ns, us, ms, and s
```

lesam:
@candrewlee14 @samhld the horizontal scroll is actually really nice for wide tables.

Here's a proposal: let's bring back the 'column' format from 1.x for this PR. I'd like to push `table` format off to a follow-up PR - we need the column format for command line tools and backward-compatibility anyway. I'd call `table` something like `column-pretty`, but I'd like to really review it in detail from a product perspective, separately from the base functionality of getting InfluxQL back in the CLI.

But the table / column-pretty format could be really nice.

DStrand1:
Don't have much else to add, code looks good and everything I could think to test works great! I'll leave it to @lesam to give it another look-through

lesam (commented):
See comments.

clients/v1_repl/v1_repl.go
```diff
@@ -0,0 +1,858 @@
...
+}
+
+func DefaultPersistentQueryParams() PersistentQueryParams {
+	return PersistentQueryParams{
+		Format:    ColumnFormat,
+		Precision: "ns",
+	}
+}
+
+func (c *Client) readHistory() []string {
```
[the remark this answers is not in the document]
candrewlee14: I pulled out minimal code for that, but I'd be happy to delete here if you want since it's still in the autocompletion PR #392.

clients/v1_repl/v1_repl.go
```diff
@@ -0,0 +1,858 @@
...
+	insertRgx := regexp.MustCompile(insertRegex)
+	insertIntoRgx := regexp.MustCompile(insertIntoRegex)
+	insertMatches := reSubMatchMap(insertRgx, cmd)
+	insertIntoMatches := reSubMatchMap(insertIntoRgx, cmd)
+	if insertIntoMatches != nil {
+		db = getIdentFromMatches(insertIntoMatches, "db")
+		rp = getIdentFromMatches(insertIntoMatches, "rp")
+		point = getIdentFromMatches(insertIntoMatches, "point")
+		if db != "" && rp == "" {
+			defaultRp, err := c.getDefaultRetentionPolicy(context.Background(), db)
```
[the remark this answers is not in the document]
candrewlee14: In 2.x we write the line protocol to a bucket, so in order to find the correct bucket to write to, I need to reverse search the default retention policy from the DBRPs because there is no actual "INSERT" command implemented at the 1.x query endpoint.

I don't know of a way to do that (db,rp)->bucket mapping without querying for the default rp if it's empty. Any ideas?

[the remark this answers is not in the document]
candrewlee14: Oh, yep, the reason I was doing it was because I was using the 2.x write endpoint and trying to generate a (db,rp)->bucket mapping when I didn't realize there was a 1.x legacy write endpoint right there. I'll take this out.

clients/v1_repl/v1_repl.go
```diff
@@ -0,0 +1,858 @@
...
+	buf := bytes.Buffer{}
+	gzw := gzip.NewWriter(&buf)
+
+	_, err := gzw.Write([]byte(point))
+	gzw.Close()
+	if err != nil {
+		color.Red("Failed to gzip points")
+		return
+	}
+	bucketID, err := c.getBucketIdForDBRP(db, rp)
```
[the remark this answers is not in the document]
candrewlee14: Wow, I somehow missed that endpoint in the docs. That'll make things a lot simpler here!

clients/v1_repl/v1_repl.go
```diff
@@ -0,0 +1,858 @@
...
+func (c *Client) query(ctx context.Context, query string) (string, error) {
+	res := c.GetLegacyQuery(ctx).
+		U(c.Username).
+		P(c.Password).
+		Db(c.Database).
+		Q(query).
+		Rp(c.RetentionPolicy).
+		Accept("application/json")
+	if c.Precision != "rfc3339" && c.Precision != "" {
+		res.Epoch(c.Precision)
```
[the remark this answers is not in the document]
candrewlee14: Good call.

clients/v1_repl/v1_repl.go
```diff
@@ -0,0 +1,759 @@
...
+	var point string
+	insertRgx := regexp.MustCompile(insertRegex)
+	insertIntoRgx := regexp.MustCompile(insertIntoRegex)
+	insertMatches := reSubMatchMap(insertRgx, cmd)
+	insertIntoMatches := reSubMatchMap(insertIntoRgx, cmd)
+	if insertIntoMatches != nil {
+		db = getIdentFromMatches(insertIntoMatches, "db")
+		rp = getIdentFromMatches(insertIntoMatches, "rp")
+		point = getIdentFromMatches(insertIntoMatches, "point")
+	} else if !insertIntoRgx.Match([]byte(cmd)) && insertMatches != nil {
```
lesam, line 197:
```suggestion
	} else if insertMatches != nil {
```

clients/v1_repl/v1_repl.go
```diff
@@ -0,0 +1,759 @@
...
+		db = c.Database
+		rp = c.RetentionPolicy
+		point = getIdentFromMatches(insertMatches, "point")
```
lesam, line 200: I think I like the little parser we had before better, but this works so 👍 .

lesam: Could you extract this parsing logic (going from `cmd` to `db, rp, point`) into a function, and write a unit test for it? That way we can make sure the various things we discussed continue to work:
* whitespace
* case insensitive

clients/v1_repl/v1_repl.go
```diff
@@ -0,0 +1,759 @@
...
+type FormatFunc func(api.InfluxqlJsonResponse)
+
+var (
+	CsvFormat    FormatType = "csv"
+	JsonFormat   FormatType = "json"
+	ColumnFormat FormatType = "column"
+)
+
+func (c *Client) runAndShowQuery(query string) {
+	// TODO: guide users trying to use deprecated InfluxQL queries
```
lesam, line 246: Is there an issue for this? I like putting issue links right into the TODO comment so it is clear when reading the code whether the TODO is stale or still live.

clients/v1_repl/v1_repl.go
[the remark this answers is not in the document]
lesam: Doesn't seem resolved, so I un-resolved it.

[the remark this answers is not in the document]
lesam: Also this doesn't actually work for setting the precision (tested manually).

I think you want this:
```suggestion
		res = res.Epoch(c.Precision)
```

lesam (commented):
See comments

clients/v1_repl/v1_repl.go
candrewlee14: Made an issue https://github.com/influxdata/influx-cli/issues/397

clients/v1_repl/v1_repl.go
```diff
@@ -0,0 +1,858 @@
...
+		<UP ARROW>    previous command
+		<DOWN ARROW>  next command`)
+}
+
+func (c *Client) settings() {
+	w := new(tabwriter.Writer)
+	w.Init(os.Stdout, 0, 1, 1, ' ', 0)
+	fmt.Fprintln(w, "Setting\tValue")
+	fmt.Fprintln(w, "--------\t--------")
+	fmt.Fprintf(w, "Username\t%s\n", c.Username)
```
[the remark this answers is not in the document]
candrewlee14: Removed from this PR, added a separate PR here #396

clients/v1_repl/v1_repl.go
candrewlee14: Added a unit test file.

clients/v1_repl/v1_repl.go
```diff
@@ -0,0 +1,783 @@
...
+	displayFunc := displayMap[c.Format]
+	displayFunc(response)
+}
+
+func (c *Client) help() {
+	fmt.Println(`Usage:
+        pretty                toggles pretty print for the json format
+        use <db_name>         sets current database
+        format <format>       specifies the format of the server responses: json, csv, column
+        precision <format>    specifies the format of the timestamp: h, m, s, ms, u or ns
```
lesam, line 299: We should print rfc3339 here

clients/v1_repl/v1_repl.go
```diff
@@ -0,0 +1,783 @@
...
+		color.Red("Expected a precision [rfc3339, ns, u, ms, s, m, or h]")
+		return
+	}
+	precision := args[1]
+	switch precision {
+	case "rfc3339", "ns", "u", "µ", "ms", "s", "m", "h":
+		c.Precision = precision
+	default:
+		color.HiRed("Unimplemented precision %q, keeping %s precision.", precision, c.Precision)
+		color.HiBlack("Choose a precision from [ns, u, ms, s, m, or h]")
```
lesam, line 381: We should print rfc3339 here

lesam (approved):
🎉

Issue #391, opened 2022-06-01
Add autocompletion to REPL


candrewlee14:
This is a placeholder issue that will be implemented in a separate PR because of PR https://github.com/influxdata/influx-cli/pull/386 getting too large.

PR #392, opened 2022-06-01
Add REPL autocompletion & go-prompt

candrewlee14:
I took this out of #386 because the PR was getting too big, so this can be reviewed separately.
This brings back the history scrolling with arrow keys, autocompletion, and <ctrl+d> exit keybinding all through go-prompt.
Closing #391


clients/v1_repl/v1_repl.go
```diff
@@ -297,6 +453,44 @@ func (c *Client) runAndShowQuery(query string) {
 	displayFunc(response)
 }
 
+// This function generates the prompt autocompletions
+func (c *Client) completer(d prompt.Document) []prompt.Suggest {
```
lesam, line 457: Nitpick - is it possibly to arrange that `use` comes first in the completions if the user hasn't picked a database, and `SELECT` and `INSERT` are the first completions if the user has picked a database (followed by the non-query functions like `clear`, `format`, etc)

clients/v1_repl/v1_repl.go
```diff
@@ -75,8 +119,9 @@ func (c *Client) clear(cmd string) {
 
 func DefaultPersistentQueryParams() PersistentQueryParams {
 	return PersistentQueryParams{
-		Format:    ColumnFormat,
-		Precision: "ns",
+		Format:       ColumnFormat,
+		Precision:    "ns",
+		historyLimit: 100,
```
lesam, line 124: We have this in two places, it should be extracted to a constant.

clients/v1_repl/v1_repl.go
lesam: Also after `SELECT` it would be cool to turn off the suggestions until after the `FROM` - it's difficult to suggest field names and the `ALL` `ANY` etc get distracting.

lesam: We already have the `influxql` repo that has a parser implemented, I wonder how hard it would be to add a feature there to return an intermediate parse result that hints at the actual possible completions instead of giving all the possible InfluxQL keywords even when they are not relevant?

clients/v1_repl/v1_repl.go
```diff
@@ -90,18 +135,128 @@ func (c *Client) Create(ctx context.Context) error {
...
+		}
 	}
+	if homeDir := os.Getenv("HOME"); homeDir != "" {
+		historyDir = homeDir
+	}
+	var history []string
+	if historyDir != "" {
+		c.historyFilePath = filepath.Join(historyDir, ".influx_history")
+		history = c.readHistory()
+		c.rewriteHistoryFile(history)
```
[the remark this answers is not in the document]
candrewlee14: For each command, we write to the history file, but there was no place where the file would be bounded in size. We'd have to read the entire file and pull out the last 100 entries, but it would only ever get bigger.
I figured that the start of the REPL was a good time to truncate the file back down to only the latest 100 commands.

clients/v1_repl/v1_repl.go
```diff
@@ -111,6 +266,7 @@ func (c *Client) executor(cmd string) {
 	if cmd == "" {
 		return
 	}
+	defer c.writeCommandToHistory(cmd)
```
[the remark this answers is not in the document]
candrewlee14: Yeah, I didn't think that having `quit`/`exit` in the command history would be useful for the user.
If you think it would be, I can just have it write before exiting, let me know.

clients/v1_repl/v1_repl.go
```diff
@@ -39,11 +42,52 @@ type PersistentQueryParams struct {
...
+			}
+			return history
+		}
+	}
+	return []string{}
+}
+
+func (c *Client) rewriteHistoryFile(history []string) {
+	if c.historyFilePath != "" {
+		if historyFile, err := os.OpenFile(c.historyFilePath, os.O_TRUNC|os.O_CREATE|os.O_WRONLY, 0644); err == nil {
```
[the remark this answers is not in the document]
candrewlee14: Oh, totally. Didn't realize that `os.Create` truncated the file if it exists, thanks for the suggestion.

clients/v1_repl/v1_repl.go
candrewlee14: Looks like these lines here implement the logic of the parser: https://github.com/influxdata/influxql/blob/1ba470371ec093d57a726b143fe6ccbacf1b452b/parse_tree.go#L102-L235

I'm not sure if I can use that code directly, but I can structure the autocompletion with the same logic if that makes sense.

clients/v1_repl/v1_repl.go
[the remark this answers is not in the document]
lesam: That's worth a comment I think.

clients/v1_repl/v1_repl.go
lesam: I'm a bit worried about maintenance here - I don't want to end up with two implementations of the InfluxQL parser.

candrewlee14: I added made a new commit to the PR that handles autocompletion a much more extensible way. You may see it as another implementation of the influxql parser, but the builtin REPL commands combined with the deprecated influxql commands in 2.x make it different enough to me that it warrants this change. Let me know what you think.

candrewlee14: Also, I've got an idea for helping guide users through deprecated queries. I can add a feature to the go-prompt library to allow grayed-out or yellow options signifying deprecated commands. Those options wouldn't be selectable, and they would always appear after non-deprecated options. What do you think of that idea in terms of resolving #397 ?

clients/v1_repl/autocomplete.go
```diff
@@ -0,0 +1,181 @@
...
+			return map[string]SuggestNode{
+				"column": {},
+				"csv":    {},
+				"json":   {},
+			}, rem
+		}},
+		"SELECT":      {subsuggestFn: c.suggestSelect},
+		"INSERT":      {},
+		"INSERT INTO": {},
+		"DELETE":      {},
```
lesam, line 118: `DELETE ... FROM <measurement>` and `DROP MEASUREMENT <measurement> ` could also suggest measurements

clients/v1_repl/v1_repl.go
lesam: I think I'd prefer to just not complete the non-implemented InfluxQL at all - it's possible we would implement more of the old InfluxQL spec in the future, so I don't want to be too aggressive. `https://docs.influxdata.com/influxdb/v2.2/query-data/influxql/#` gives the implemented vs. non-implemented subset.

Also it would be nice (though not critical) if the REPL was backward-compatible to old 1.x databases.

Maybe we could enhance the messages from the database itself for old queries? It is already parsing the influxql statements, it would be easy enough to say something like `CREATE DATABASE is a 1.x DML statement that is not supported` direct from the 2.x API, and then the client can just report the error it gets back.

clients/v1_repl/autocomplete.go
```diff
@@ -0,0 +1,211 @@
...
+				"FIELD KEYS":            {},
+				// "GRANTS":                {},
+				// "MEASUREMENT CARDINALITY":       {},
+				"MEASUREMENT EXACT CARDINALITY": {},
+				"MEASUREMENTS":                  {},
+				// "QUERIES":                 {},
+				// "RETENTION POLICIES":      {},
+				"SERIES": {},
+				// "SERIES CARDINALITY": {},
+				"SERIES EXACT CARDINALITY": {},
```
lesam, line 164: Does this work against 2.x / cloud?

clients/v1_repl/autocomplete.go
```diff
@@ -0,0 +1,211 @@
...
+				// "STATS":                   {},
+				// "SUBSCRIPTIONS":           {},
+				"TAG KEY CARDINALITY":       {},
+				"TAG KEY EXACT CARDINALITY": {},
+				"TAG KEYS":                  {},
+				"TAG VALUES":                {},
+				"USERS":                     {},
+			}, rem
+		}},
+		"CREATE": {subsuggestFn: func(rem string) (map[string]SuggestNode, string) {
```
lesam, line 176: I don't think any of these `CREATE` statements work against 2.x - I think we should remove suggestions for `CREATE`

clients/v1_repl/autocomplete.go
```diff
@@ -0,0 +1,211 @@
...
+				"USER":             {},
+				"RETENTION POLICY": {},
+				"SUBSCRIPTION":     {},
+			}, rem
+		}},
+		"DROP": {subsuggestFn: func(rem string) (map[string]SuggestNode, string) {
+			return map[string]SuggestNode{
+				"CONTINUOUS QUERY": {},
+				"DATABASE":         {subsuggestFn: c.suggestDropDatabase},
+				"MEASUREMENT":      {subsuggestFn: c.suggestDropMeasurement},
```
lesam, line 189: Only `DROP MEASUREMENT` is supported at least for now, so I think we should only suggest that one.

clients/v1_repl/autocomplete.go
candrewlee14: Yeah, at least for 2.x OSS, `SHOW SERIES EXACT CARDINALITY` is implemented while `SHOW SERIES CARDINALITY` is not.

clients/v1_shell/autocomplete.go
```diff
@@ -0,0 +1,203 @@
...
+			for _, m := range c.Measurements {
+				s["\""+m+"\""] = SuggestNode{Description: fmt.Sprintf("Measurement on \"%s\".\"%s\"", c.Database, c.RetentionPolicy)}
+			}
+		}
+		return s, getIdentFromMatches(matches, "from_clause")
+	}
+	return s, remainder
+}
+
+func (c *Client) suggestSelect(remainder string) (map[string]SuggestNode, string) {
```
lesam, line 50: It could be really cool to complete fields names eventually - not for this PR though. Needs some UX thought.

lesam (approved):

PR #393, opened 2022-06-03
add pretty table format to REPL

candrewlee14:
This PR separates out the new `table` format from PR #386.


clients/v1_shell/v1_shell.go
```diff
@@ -51,7 +52,7 @@ type PersistentQueryParams struct {
 
 func DefaultPersistentQueryParams() PersistentQueryParams {
 	return PersistentQueryParams{
-		Format:       ColumnFormat,
+		Format:       TableFormat,
```
lesam, line 55: Do we have a `-execute` flag yet? When/if we add one (to run in a script) we should make `ColumnFormat` the default for that.

clients/v1_shell/v1_shell.go
```diff
@@ -621,6 +624,22 @@ func (c *Client) outputColumns(response api.InfluxqlJsonResponse) {
 	writer.Flush()
 }
 
+func (c *Client) outputTable(response api.InfluxqlJsonResponse) {
+	for _, res := range response.GetResults() {
```
lesam, line 628: Should we create a new window for every table? For queries that create a bunch of small tables this seems awkward.

clients/v1_shell/table_model.go
```diff
@@ -0,0 +1,150 @@
...
+
+type Model struct {
+	rows        []table.Row
+	allCols     []table.Column
+	colNames    []string
+	simpleTable table.Model
+	totalWidth  int
+}
+
+func NewModel(res api.InfluxqlJsonResponseSeries) Model {
```
lesam, line 28: How does the `>` at the end of a wide table work? I can't see where that gets set up.

lesam (commented):
See comments / questions

clients/v1_shell/table_model.go
candrewlee14: Happens in the bubble-table library, it's a new feature upstream that I suggested.

clients/v1_shell/v1_shell.go
candrewlee14: No flag yet, I can create an issue for it.

clients/v1_shell/v1_shell.go
candrewlee14: When the table is small enough to fit on screen with no scrolling, there is no interactive view and it prints all the tables in one go. Here's an example:
Screen Shot 2022-06-23 at 11 25 14 AM
What do you think of that behavior?

lesam (approved):
Screenshot 2022-06-24 at 10 35 27 AM


This seems backward - the right-most thing should be the 'biggest' object. Also I would call `Result` as (`Statement`, or `Table Set`) and `Series` as (`Group`, or `Table`)

lesam:
@samhld see comment above.

samhld:
@lesam yeah agreed. Can we merge this bad boy and get to that afterward?

PR #398, opened 2022-06-10
fix: users and orgs permissions should not be scoped under an org

### LINKED ISSUE #293: `auth create` creates improperly-scoped permissions for `--<action>-orgs` and `--<action>-users`

jeffreyssmith2nd:
Closes #293 

Users and Orgs permissions shouldn't be scoped under a specific organization, this fixes that.

### DIFF
--- a/clients/auth/auth.go
+++ b/clients/auth/auth.go
@@ -444,6 +444,9 @@ func (c Client) printAuth(opts printParams) error {
 }
 
 func makePermResource(permType string, id string, orgId string) api.PermissionResource {
+	if permType == "orgs" || permType == "users" {
+		orgId = ""
+	}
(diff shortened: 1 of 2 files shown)


davidby-influx (approved):
LGTM

Issue #378, opened 2022-05-18
Improve parsing of token flag


samhld:
A [community issue](https://influxcommunity.slack.com/archives/CH8RV8PK5/p1652864367351869) was raised that when a token starts with `-`, the `--token` flag when written without an `=` (`--token <arg>` vs `--token=<value>`), the CLI trips up.

We should allow for all valid tokens to be passed to either way of writing the flag. This issue is likely to manifest 1/64th of the time users use the former version of the flag.

Issue #376, opened 2022-05-04
Token starting with `-` is interpreted as flag


LarsStegman:
I have a token that starts with a `-`, but this is interpreted as a flag when I pass it as a value for `-t`. Even surrounding the token with `"` causes a failure.

## Input
 `influx config create -n targetflux -u "http://localhost:8087" -t "-therestofmytoken" -o "organisation" -a`

## Result
`Error: unknown command "organisation" for "create"`

PR #399, opened 2022-06-14
Replace token flags with = to prevent bad parsing of leading dash in token

candrewlee14:
This is a pretty awkward problem, and it stems from the fact that those wrapping quotes are parsed away when Go gets the program's arguments for `os.Args`.

I ran `dlv debug ./cmd/influx -- config create -n targetflux -u "http://localhost:8087" -o "organization" -t "-therestofmytoken"`
And this is the `os.Args` value which `urfave/cli` uses directly.
```
(dlv) p os.Args
[]string len: 11, cap: 11, [
        "...+2 more",
        "config",
        "create",
        "-n",
        "targetflux",
        "-u",
        "http://localhost:8087",
        "-o",
        "organization",
        "-t",
        "-therestofmytoken",      <--- no wrapping quotes
]
```
It seems like a bespoke method specifically for replacing `-t "-FOO-TOKEN"` with `-t=-FOO-TOKEN` in `os.Args` before passing into `urfave/cli` makes the most sense here.

Closing #376 and #378

### DIFF
--- a/cmd/influx/main.go
+++ b/cmd/influx/main.go
@@ -3,6 +3,7 @@ package main
 import (
 	"fmt"
 	"os"
+	"strings"
 	"time"
 
 	"github.com/influxdata/influx-cli/v2/pkg/cli/middleware"
@@ -91,8 +92,31 @@ func newApp() cli.App {
 	}
 }
 
+// This replaces `-t "-FOO-TOKEN"` with `-t=-FOO-TOKEN` in os.Args
+// This is necessary to do because the command line arg:
+//  `-t "-FOO-TOKEN"`` will be parsed as two separate flags instead of a flag and token value.
+func replaceTokenArg() {
+	for i, arg := range os.Args {
+		switch arg {
+		case "--token", "-t":
+			// if last element, this will be invalid later
+			if i == len(os.Args)-1 {
+				break
+			}
+			os.Args[i] = strings.Join(os.Args[i:i+2], "=")
+			// if there are 2+ elements after this
+			if len(os.Args) > i+2 {
+				os.Args = append(os.Args[:i+1], os.Args[i+2:]...)
+			} else {
+				os.Args = os.Args[:i+1]
+			}
+		}
+	}
+}
+
 func main() {
 	app := newApp()
+	replaceTokenArg()
 	if err := app.Run(os.Args); err != nil {
 		// Errors will normally be handled by cli.HandleExitCoder via ExitErrHandler set on app. Any error not implementing
 		// the cli.ExitCoder interface can be handled here.
--- a/cmd/influx/main.go
+++ b/cmd/influx/main.go
@@ -96,30 +96,22 @@ func newApp() cli.App {
 // This creates a new slice and replaces `-t "-FOO-TOKEN"` with `-t=-FOO-TOKEN`
 // This is necessary to do because the command line arg:
 //  `-t "-FOO-TOKEN"`` will be parsed as two separate flags instead of a flag and token value.
-func replaceTokenArg(args []string) []string {
+func ReplaceTokenArg(args []string) []string {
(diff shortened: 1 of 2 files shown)


cmd/influx/main.go
```diff
@@ -91,8 +92,31 @@ func newApp() cli.App {
...
+			} else {
+				os.Args = os.Args[:i+1]
+			}
+		}
+	}
+}
+
 func main() {
 	app := newApp()
+	replaceTokenArg()
```
lesam, line 119: This is clever! Can we create a new slice instead of messing with the os.Args one though (and pass that to `app.Run`?)

samhld:
Will both options still be available from the user perspective?

candrewlee14:
> Will both options still be available from the user perspective?

Yeah, it just rewrites the space-flag syntax to an equal-flag syntax for the token flag behind the scenes. User passes their arguments the same way.

lesam:
The problem would be if people write something like:

```
./influx config create -n local_2 -u http://localhost:8086/ -t -a
Active  Name    URL                     Org
        local_2 http://localhost:8086/
```

The `-a` gets interpreted as a token when it was intended as an `active` marker, which is weird.

On the whole, I think this is better than the alterative though.

Maybe we could have a stderr print like `warning: -t -a interpretted as -t=-a, consider using -t=-a syntax when tokens may start with a hyphen`.

@samhld thoughts?

samhld:
@lesam I think that works if this is the path we're taking. What was the last sentiment on just changing the token generation server-side to disallow this type of token. Do you think it's reasonable users would bring their own token starting with a hyphen? I suppose we'd have to support that even if it's an edge case, if we're going to support "bring your own token" at all.

cmd/influx/main.go
```diff
@@ -91,9 +93,42 @@ func newApp() cli.App {
...
+	copy(newArgs, args)
+	for i, arg := range newArgs {
+		switch arg {
+		case "--token", "-t":
+			// if last element, this will be invalid later
+			if i == len(args)-1 {
+				break
+			}
+			if strings.HasPrefix(newArgs[i+1], "-") {
+				color.HiYellow("warning: %s %s interpreted as %s=%s, consider using %s=%s syntax when tokens start with a hyphen",
```
lesam, line 110: Can you use explicit argument indexing in the format string here?

cmd/influx/main.go
```diff
@@ -91,9 +93,42 @@ func newApp() cli.App {
...
+					newArgs[i], newArgs[i+1],
+					newArgs[i], newArgs[i+1],
+					newArgs[i], newArgs[i+1],
+				)
+			}
+			newArgs[i] = strings.Join(newArgs[i:i+2], "=")
+			// if there are 2+ elements after this
+			if len(newArgs) > i+2 {
+				newArgs = append(newArgs[:i+1], newArgs[i+2:]...)
```
lesam, line 119: I don't think you need the 'if' here - `newArgs` would be an empty slice if `len(newArgs) == i+2`

cmd/influx/main.go
```diff
@@ -91,9 +93,42 @@ func newApp() cli.App {
 	}
 }
 
+// This creates a new slice and replaces `-t "-FOO-TOKEN"` with `-t=-FOO-TOKEN`
+// This is necessary to do because the command line arg:
+//  `-t "-FOO-TOKEN"`` will be parsed as two separate flags instead of a flag and token value.
+func replaceTokenArg(args []string) []string {
```
lesam, line 99: Can you write a unit test for `replaceTokenArg` ?

lesam (changes requested):

cmd/influx/main.go
candrewlee14: TIL!

lesam (approved):

Issue #231, opened 2021-08-20
Allow setting membership type in `influx org members add`


danxmoran:
`influx org members add` is hard-coded to add new users to an org as a "member". We should add a way for users to be added as an "owner" (the API already supports this, so we need to add a toggle to the CLI).

`influx org members list` and `influx org members remove` may also need to be updated to account for owners as well as members.

lesam:
Until this is implemented, you can get effectively the same thing by granting a user an all-access token on the org, since users inherit the permissions of all the tokens that the own:

```
influx auth create --all-access -u <userNeedingPermission> -o <theorg>
```

kuzyn:
Is it possible to link on how to do this with the API?

glacieruk:
What release is this in?

samhld:
@glacieruk it hasn't been released yet. Next release!

PR #402, opened 2022-06-23
Allow setting membership type in influx org members add

candrewlee14:
This allows listing, adding, and removing owners from an org with `influx org members <command>`.
- Listing will list owners and members.
- Adding will add member by default, or owner with `--owner` flag.
- Removing will remove owners and members, no flag needed.

Closing #231

### DIFF
--- a/clients/org/org_members_test.go
+++ b/clients/org/org_members_test.go
@@ -159,13 +159,12 @@ func TestClient_ListMembers(t *testing.T) {
 	t.Parallel()
 
 	testCases := []struct {
-		name                     string
-		params                   org.ListMemberParams
-		defaultOrgName           string
-		registerOrgExpectations  func(*testing.T, *mock.MockOrganizationsApi)
-		registerUserExpectations func(*testing.T, *mock.MockUsersApi)
-		expectedOut              []string
-		expectedErr              string
+		name                    string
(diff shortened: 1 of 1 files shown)


cmd/influx/org_members.go
```diff
@@ -45,6 +45,11 @@ func newOrgMembersAddCmd() cli.Command {
 				EnvVar:      "INFLUX_ORG_ID",
 				Destination: &params.OrgID,
 			},
+			&cli.BoolFlag{
+				Name:        "owner",
```
DStrand1, line 49: maybe allow `o` here as well for convenience?

samhld:
Seems reasonable to me! If we want to fully satisfy the original issue, we could add a separate hint to point the user at the UI. I imagine most users would know about it but I could see them not. The question would be, can we point them at the exact address of where the UI is? Any discovery tooling to point them exactly where to go to see the UI?

DStrand1 (approved):
LGTM! Just one small suggestion for the CLI command usage

cmd/influx/org_members.go
candrewlee14: I wondered if users might accidentally use `-o` thinking they were passing an org flag, but actually set the user as owner. What do you think @samhld ?

Issue #407, opened 2022-06-24
Influx CLI: Add ability to create a Task linked to a Script


brettbuddin:
When creating a task, we can either provide flux that the task executes or a script ID with optional parameters.

https://github.com/influxdata/influx-cli/blob/d3e0efb560b6272e18d6436cc25e52f7e7036a17/cmd/influx/task.go#L32
https://github.com/influxdata/influx-cli/blob/d3e0efb560b6272e18d6436cc25e52f7e7036a17/clients/task/task.go#L70

PR #413, opened 2022-06-24
feat: creating tasks with invokable scripts and more create params

Xenonic:
This PR allows users to create tasks with script IDs instead of Flux in the CLI. The following parameters are now exposed when creating a task:
 - name
 - every
 - cron
 - scriptID
 - scriptParameters

Only the last two are related to scripts, but if a name and schedule aren't provided the command will fail when trying to create a task (https://github.com/influxdata/idpe/blob/47a5a53d9d00313603ef6f3b5991f3a68122ac4f/task/task.go#L453)

Must merge https://github.com/influxdata/openapi/pull/349 and update the submodule before this PR can be merged!

Closes https://github.com/influxdata/influx-cli/issues/407


Xenonic:
Superseded by https://github.com/influxdata/influx-cli/pull/432

PR #417, opened 2022-06-28
feat: allow deleting replications with remotes

DStrand1:
https://github.com/influxdata/influxdb/pull/23493 removed replication configs deleting automatically when remotes are deleted. This PR reimplements that feature, but only if you pass the `--delete-replications` flag to the `influx remote delete` command

### DIFF
--- a/cmd/influx/remote.go
+++ b/cmd/influx/remote.go
@@ -1,7 +1,10 @@
 package main
 
 import (
+	"fmt"
+
 	"github.com/influxdata/influx-cli/v2/clients/remote"
+	"github.com/influxdata/influx-cli/v2/clients/replication"
 	"github.com/influxdata/influx-cli/v2/pkg/cli/middleware"
 	"github.com/urfave/cli"
 )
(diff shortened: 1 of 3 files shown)


cmd/influx/remote.go
```diff
@@ -94,16 +98,40 @@ func newRemoteDeleteCmd() cli.Command {
...
 
 			client := remote.Client{
-				CLI:                  getCLI(ctx),
+				CLI:                  cli,
 				RemoteConnectionsApi: api.RemoteConnectionsApi,
 			}
 
-			return client.Delete(getContext(ctx), remoteID)
+			if deleteReplications {
+				conn, err := client.Get(context, remoteID)
```
DStrand1, line 118: I don't like doing this client action here and pulling an API struct into `cmd`, but the alternative is doing replication things in `remotes` dir, or the other way around, so I think this is the best way to handle it and keep these two as separated as possible

jeffreyssmith2nd (approved):

PR #418, opened 2022-06-29
feat: add username and password login

jeffreyssmith2nd:
This PR adds the ability for a user to login to InfluxDB using a username and password rather than a token. A user can specify both the username and password to be saved and used for all requests. Or, a user can specify just the username and they will be prompted for the password each time.
*Note*: this feature is OSS only


clients/signin/signin.go
```diff
@@ -0,0 +1,56 @@
...
+
+	cfg := api.NewAPIConfig(params)
+	client := api.NewAPIClient(cfg)
+	ctx = context.WithValue(ctx, api.ContextBasicAuth, api.BasicAuth{
+		UserName: username,
+		Password: password,
+	})
+	res, err := client.SigninApi.PostSignin(ctx).ExecuteWithHttpInfo()
+	if err != nil {
+		return "", err
```
candrewlee14, line 47: Maybe if an error happens here, you could again print to the user that this is only supported for OSS, in case that clarifies their problem. What does the error look like when you try to make this request to cloud?

clients/signin/signin.go
```diff
@@ -0,0 +1,56 @@
...
+	}
+
+	cookies := res.Cookies()
+	if len(cookies) != 1 {
```
candrewlee14, line 51: Would cookies ever have a length greater than 1? Should that be an error if it happens?

candrewlee14 (commented):
Looks good, just a couple questions.

clients/signin/signin.go
jeffreyssmith2nd: Only if we changed something on the influxdb side of things. I'm fine with either erroring or just taking the first cookie if there's multiples. I went this route because it might highlight unexpected behavior coming from influxdb

clients/signin/signin.go
jeffreyssmith2nd: It simply 404s because cloud tries to do some redirect stuff that we don't follow. I can wrap this error with a suggestion to make sure its not running against cloud.

candrewlee14 (approved):
LGTM

PR #420, opened 2022-06-30
fix: improve alignment for strings and numbers in v1 shell

### LINKED ISSUE #405: [v1 shell] EXPLAIN ANALYZE formatted poorly in table format
Assuming this will persist to the new table format (I'm using the one from the `add-repl-table-format` branch), the text looks wonky in a table. 

Screen Shot 2022-06-23 at 10 18 08 PM

I think we have two options:
- format it better in the table cell
- error and tell the user to use the column format for explain plans

candrewlee14:
- improves string formatting so that EXPLAIN ANALYZE format looks correct
- adds right alignment for number types, and left alignment for string (no more quotes as well)
- adds `scientific` command to toggle decimal vs. scientific numeric format when using table format.
- enter altscreen when running table interactive mode to improve scroll appearance. 

Closes #405

### DIFF
--- a/clients/v1_shell/autocomplete.go
+++ b/clients/v1_shell/autocomplete.go
@@ -118,8 +118,9 @@ func newSugNodeFn(subsugs ...string) subsuggestFnType {
 func (c *Client) completer(d prompt.Document) []prompt.Suggest {
 	// the commented-out lines are unsupported in 2.x
 	suggestions := map[string]SuggestNode{
-		"use":    {Description: "Set current database", subsuggestFn: c.suggestUse},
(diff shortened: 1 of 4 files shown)


candrewlee14:
Before:
Before
After:
Screen Shot 2022-06-30 at 2 20 09 PM

candrewlee14:
New formatting for numbers & strings (now colored yellow with no quotes).
Toggle number format to scientific or back to decimal with the `scientific` command.
Screen Shot 2022-06-30 at 3 26 16 PM

jeffreyssmith2nd (approved):

PR #428, opened 2022-07-20
feat: added tag stripping step to openapi generation to fix codegen

Xenonic:
Upstream versions of the openapi submodule currently break code generation, since some operations have multiple tags, which causes multiple duplicate and conflicting versions to be generated. This PR strips out the newly introduced grouping tags to restore the previous codegen behavior.

More details: https://github.com/influxdata/openapi/pull/445


etc/stripGroupTags.py
```diff
@@ -0,0 +1,15 @@
+import sys
+import re
+
+blacklist = [
+    "Data I/O endpoints",
+    "Security and access endpoints",
+    "System information endpoints"
+]
```
glinton, line 8: Maintaining this list will likely be a painpoint moving forward. Can you block the parent `Tags` field and it's contents altogether? Might have to utilize proper yaml parsing, but the more hands-off it is, the better.

etc/generate-openapi.sh
```diff
@@ -28,18 +28,22 @@ docker run --rm -it -u "$(id -u):$(id -g)" \
   --outfile /api/cli-extras.gen.yml \
   --type yaml
 
+# Strip certain tags to prevent duplicated and conflicting codegen.
+python3 ./etc/stripGroupTags.py ./api/cli.gen.yml > ./api/cli-stripped.gen.yml
```
glinton, line 32: For portability, make this a docker command so `python3` isn't required to be in someone's path

etc/generate-openapi.sh
```diff
@@ -28,18 +28,22 @@ docker run --rm -it -u "$(id -u):$(id -g)" \
...
   generate \
   -g go \
-  -i /api/cli.gen.yml \
+  -i /api/cli-stripped.gen.yml \
   -o /api \
   -t /api/templates \
   --additional-properties packageName=api,enumClassPrefix=true,generateInterfaces=true
 
+rm ./api/cli-stripped.gen.yml
+
```
glinton, line 46: This is fine, but to follow the current pattern, just add the new file to the `.gitignore`

etc/stripGroupTags.py
Xenonic: I initially tried removing all tags, but this causes the API stubs to be thrown in a generic interface, instead of an interface such as `Tasks`. I wasn't able to find a way to group apis together without using the tag system either.

glinton: So strange.. I would expect from your former error:
```
api/api_security_and_access_endpoints.gen.go:618:6: ApiGetAuthorizationsRequest redeclared in this block
        api/api_authorizations.gen.go:259:6: other declaration of ApiGetAuthorizationsRequest
```
that it didn't generate interfaces nor structs that implement them, but flat functions grouped into named files. Seems if they use tags to create an interface, they should preserve that thought and make functions (like `ApiGetAuthorizationsRequest` be memeber functions of some tag-named-ish struct. What uses the generated interface(s) and would it still work despite it being too coupled?

etc/generate-openapi.sh
Xenonic: fixed

etc/generate-openapi.sh
Xenonic: fixed

etc/stripGroupTags.py
Xenonic: The generated interfaces are used all throughout the codebase, not sure what you mean with `would it still work despite it being too coupled?`

glinton (approved):
:tada:

DStrand1 (approved):

PR #439, opened 2022-08-16
Clarify difference in virtual vs physical dbrps when listing

candrewlee14:
- Add new help info about virtual DBRPs for `influx v1 dbrp list -h` 
image

- Split `influx v1 dbrp list` into physical and virtual DBRP tables
command line screenshot


sanderson (approved):

candrewlee14:
@sanderson I'm still handling tests, which are based on the previous output format of v1 dbrp list. Should have those cleaned up tonight or tomorrow morning.

PR #440, opened 2022-08-17
feat: replication bucket name

jeffreyssmith2nd:
Replications will now support passing a bucket-name as well as a bucket-id. This is to provide full support for replication to 1.x installations. Fixes https://github.com/influxdata/influxdb/issues/23550

OpenAPI PR: https://github.com/influxdata/openapi/pull/478

### DIFF
--- a/clients/bucket/create.go
+++ b/clients/bucket/create.go
@@ -34,10 +34,11 @@ func (c Client) Create(ctx context.Context, params *BucketsCreateParams) error {
 		return err
 	}
 
+	var rr []api.RetentionRule
 	reqBody := api.PostBucketRequest{
 		OrgID:          orgId,
 		Name:           params.Name,
-		RetentionRules: []api.RetentionRule{},
+		RetentionRules: &rr,
 		SchemaType:     &params.SchemaType,
 	}
 	if params.Description != "" {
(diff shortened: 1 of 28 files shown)


cmd/influx/replication.go
```diff
@@ -66,9 +66,13 @@ func newReplicationCreateCmd() cli.Command {
 			&cli.StringFlag{
 				Name:        "remote-bucket-id",
 				Usage:       "ID of remote bucket data should be replicated to",
-				Required:    true,
 				Destination: &params.RemoteBucketID,
 			},
+			&cli.StringFlag{
+				Name:        "remote-bucket-name",
```
sanderson, line 72: In all other commands that take a bucket name or bucket ID, the flags are `--bucket` and `--bucket-id`. Should this flag just be `--remote-bucket`?

clients/bucket/create.go
```diff
@@ -34,10 +34,11 @@ func (c Client) Create(ctx context.Context, params *BucketsCreateParams) error {
 		return err
 	}
 
+	var rr []api.RetentionRule
 	reqBody := api.PostBucketRequest{
 		OrgID:          orgId,
 		Name:           params.Name,
-		RetentionRules: []api.RetentionRule{},
+		RetentionRules: &rr,
```
lesam, line 41: Why this change?

cmd/influx/main.go
```diff
@@ -96,7 +96,8 @@ func newApp() cli.App {
 
 // This creates a new slice and replaces `-t "-FOO-TOKEN"` with `-t=-FOO-TOKEN`
 // This is necessary to do because the command line arg:
-//  `-t "-FOO-TOKEN"`` will be parsed as two separate flags instead of a flag and token value.
+//
+//	`-t "-FOO-TOKEN"`` will be parsed as two separate flags instead of a flag and token value.
```
lesam, line 100: I think this comment was more clear before

lesam (commented):
LGTM, a few minor comments and needs openapi updated.

cmd/influx/main.go
jeffreyssmith2nd: Seems like `gofmt` wants to force it on a newline if it has the leading tab. I removed that to remove the empty line.

PR #444, opened 2022-09-16
fix: fixes an error where stdin could not be used to create tasks

jeffreyssmith2nd:
Closes https://github.com/influxdata/influxdb/issues/23672


davidby-influx (approved):
LGTM.

Sort of would prefer the argument handling pulled up out of `ReadQuery` but that's not necessary for this PR.

PR #448, opened 2022-10-13
chore: fix go version in `go.mod`

DStrand1:


jeffreyssmith2nd (approved):
Thanks, I should have caught this when I did the upgrade

PR #449, opened 2022-10-18
fix: cloud expects dataType in csv files

jeffreyssmith2nd:
Closes https://github.com/influxdata/influx-cli/issues/450


gwossum (approved):
LGTM, assuming I'm correct that the bucket-schema command is not supported by OSS

PR #451, opened 2022-10-19
fix: do not require remoteOrgID for remote connection creation

DStrand1:
This PR makes the remote creation command no longer require a remote org ID. There is a warning suggesting that it is needed for Cloud or OSS, but not for 1.x.

Additionally, this PR updates the CLI to the latest OpenAPI, which came with a handful of changes, though all of the changes in this PR for that are just updating for changes and not making any user-facing changes


api/contract/cli-extras.yml
```diff
@@ -9,5 +9,5 @@ components:
     ResourceEnumOSS:
       $ref: "./openapi/src/common/schemas/Resource.yml#/properties/type"
     ResourceEnumCloud:
-      $ref: "./openapi/src/cloud/schemas/Resource.yml#/properties/type"
+      $ref: "./openapi/src/common/schemas/Resource.yml#/properties/type"
```
DStrand1, line 12: This change is required as of https://github.com/influxdata/openapi/pull/557

jeffreyssmith2nd (dismissed):

davidby-influx (approved):
LGTM

jeffreyssmith2nd (approved):

PR #456, opened 2022-11-02
chore: upgrade Go to 1.19.3 - 2.5

jdstrand:


jdstrand:
I prepared this thinking that 2.5.0 hadn't been released yet, but it has. As https://github.com/influxdata/influx-cli/pull/455 is committed, we are ready for this to be incorporated into the next 2.6.0 release. Closing.

If you need to release a 2.5.1 of `influx-cli`, it would be nice to re-open this PR and apply it.

PR #480, opened 2023-03-06
feat: Add new replication field to better show queue sync progress

jeffreyssmith2nd:


clients/auth/auth.go
```diff
@@ -12,7 +12,7 @@ import (
 
 type Client struct {
 	clients.CLI
-	api.AuthorizationsApi
+	api.AuthorizationsAPITokensApi
```
davidby-influx, line 15: Is this rename still to be done?

Issue #502, opened 2023-04-24
influx write incorrectly parses input data when --skipRowOnError is enabled


lorinescu:
I am trying to import an LP file exported from influxdb v1.8 into v2.7. When I use `--skipRowOnError` the input data is not correctly parsed. When I don't use this flag everything works as expected.
I believe the problem was introduced in v2.4.0 when `--skipRowOnError` was enabled for LP format. What I think happens is that when the `LineReader` buffer fills up the data between the last `\n` in the buffer and the end of the buffer is not read by `LineProtocolFilter`.

Example:
```
# cat bug-report-data
celery_task_duration,env=MOOO,host=foobar-batch,queue=high,service=BATCH,task=monitoring.tasks.celery_throughput_high duration=0.10149335861206055 1680843957087735646
celery_task_duration,env=MOOO,host=foobar-batch,queue=high,service=BATCH,task=monitoring.tasks.celery_throughput_high duration=0.10599446296691895 1680844070301214389
celery_task_duration,env=MOOO,host=foobar-batch,queue=high,service=BATCH,task=monitoring.tasks.celery_throughput_high duration=0.11674189567565918 1680844094394325888
celery_task_duration,env=MOOO,host=foobar-batch,queue=high,service=BATCH,task=monitoring.tasks.celery_throughput_high duration=0.12114214897155762 1680844139344799828
celery_task_duration,env=MOOO,host=foobar-batch,queue=high,service=BATCH,task=monitoring.tasks.celery_throughput_high duration=0.11209964752197266 1680844301203178190
celery_task_duration,env=MOOO,host=foobar-batch,queue=high,service=BATCH,task=monitoring.tasks.celery_throughput_high duration=0.1039726734161377 1680844347876425153
celery_task_duration,env=MOOO,host=foobar-batch,queue=high,service=BATCH,task=monitoring.tasks.celery_throughput_high duration=0.1268465518951416 1680844370106414825
celery_task_duration,env=MOOO,host=foobar-batch,queue=high,service=BATCH,task=monitoring.tasks.celery_throughput_high duration=0.12303543090820312 1680844646311648005
celery_task_duration,env=MOOO,host=foobar-batch,queue=high,service=BATCH,task=monitoring.tasks.celery_throughput_high duration=0.11537313461303711 1680844695335169551
[...]

PR #503, opened 2023-04-25
fix: handle LineReader buffer wrapparound in LineProtocolFilter (#502)

lorinescu:


pkg/csv2lp/lp_reader.go
```diff
@@ -21,24 +25,35 @@ func LineProtocolFilter(reader io.Reader) *LineProtocolFilterReader {
...
-		buf := b[0:bytesRead]
-		pts, err := models.ParsePoints(buf) // any time precision because we won't actually use this point
-		if err != nil {
-			log.Printf("invalid point on line %d: %v\n", state.LineNumber, err)
-			continue
-		} else if len(pts) == 0 { // no points on this line
-			continue
+	bytesRead, err := state.lineReader.Read(state.line)
+	if err != nil && bytesRead == 0 {
+		return 0, err
```
davidby-influx, line 36: This conditional changes the behavior; in the case where bytes are read, but an error is returned (perhaps EOF from a final line without a newline), this will continue, whereas the existing code will return the bytesRead and the error.

pkg/csv2lp/lp_reader.go
```diff
@@ -21,24 +25,35 @@ func LineProtocolFilter(reader io.Reader) *LineProtocolFilterReader {
 	lineReader.LineNumber = 1 // start counting from 1
 	return &LineProtocolFilterReader{
 		lineReader: lineReader,
+		line:       make([]byte, 4096),
```
davidby-influx, line 28: `4096` should be the `defaultBufSize` constant, not a literal

pkg/csv2lp/lp_reader.go
```diff
@@ -21,24 +25,35 @@ func LineProtocolFilter(reader io.Reader) *LineProtocolFilterReader {
...
+		wrapBuf:    make([]byte, 4096),
```
davidby-influx, line 29: `4096` should be the `defaultBufSize` constant, not a literal

pkg/csv2lp/lp_reader.go
davidby-influx: Is this a fix for a problem you found, or an inadvertent change in behavior?

pkg/csv2lp/lp_reader.go
```diff
@@ -21,24 +25,35 @@ func LineProtocolFilter(reader io.Reader) *LineProtocolFilterReader {
...
-		} else if len(pts) == 0 { // no points on this line
-			continue
+	bytesRead, err := state.lineReader.Read(state.line)
+	if err != nil && bytesRead == 0 {
+		return 0, err
+	}
+
+	//read again when we read a partial line at the end of the line reader buffer
+	if bytesRead > 0 && state.line[bytesRead-1] != '\n' {
+		wrapBytesRead, _ := state.lineReader.Read(state.wrapBuf[0:])
```
davidby-influx, line 41: This code ignores any errors from `lineReader.Read`. Errors need to be handled and returned up the stack.

pkg/csv2lp/lp_reader.go
```diff
@@ -21,24 +25,35 @@ func LineProtocolFilter(reader io.Reader) *LineProtocolFilterReader {
...
+	state.LineNumber = state.lineReader.LastLineNumber
+	pts, err := models.ParsePoints(state.line[0:bytesRead]) // any time precision because we won't actually use this point
+	if err != nil {
+		log.Printf("invalid point on line %d: %v\n", state.LineNumber, err)
+		return 0, nil
+	} else if len(pts) == 0 {
+		return 0, nil
+	}
+
+	copy(b, state.line[0:bytesRead])
```
davidby-influx, line 57: This code has an extra `copy` for every read, and sometimes two extra copies on a line wrap.  

Are those copies necessary? Would it be possible to do all the reads into the passed in buffer, `b`, calling the second read (when there is no `\n` found) `b[bytesRead:]` or something like that?

davidby-influx (changes requested):
Thank you for your PR to the `influx-cli` project!  Sorry it has taken so long to review it.

I have a few suggestions for the code which will improve readability, maintainability, and possibly efficiency.  If you would like to make the modifications, I will happily re-review.  Thanks, again!

pkg/csv2lp/lp_reader.go
davidby-influx: Note: I am arguing elsewhere that this buffer is not necessary.

pkg/csv2lp/lp_reader.go
davidby-influx: Note: I am suggesting elsewhere that this buffer is unnecessary.

PR #504, opened 2023-04-25
fix(signin): Check for all available cookies instead of the count

TheiLLeniumStudios:
Hi,

It is possible to have multiple cookies returned on a successful sign-in response, specially when it is running behind a reverse proxy etc. This PR fixes the issue by looking up the cookies that contain `influxdb` within the name (it works on both OSS and cloud) and using the value of that cookie instead of checking for cookie array size.


jeffreyssmith2nd (approved):
Thanks for the contribution! 🎉

PR #511, opened 2023-05-11
feat: init api server configs with path of host 

DonCamillo72:
init api server configs with path of host

### DIFF
--- a/api/configuration_default.go
+++ b/api/configuration_default.go
@@ -27,6 +27,15 @@ func NewAPIConfig(params ConfigParams) *Configuration {
 	apiConfig.Scheme = params.Host.Scheme
 	apiConfig.UserAgent = params.UserAgent
 	apiConfig.HTTPClient = &http.Client{Transport: clientTransport}
+
+	if params.Host.Path != "" {
+		// initialize api server configurations with path of host  url
+		apiConfig.Servers = ServerConfigurations{{URL: params.Host.Path}}
+		for key := range apiConfig.OperationServers {
+			apiConfig.OperationServers[key] = ServerConfigurations{{URL: params.Host.Path}}
+		}
+	}
+
 	if params.Token != nil && *params.Token != "" {
 		apiConfig.DefaultHeader["Authorization"] = fmt.Sprintf("Token %s", *params.Token)
 	} else if params.Cookie != nil && *params.Cookie != "" {


srebhan:
@albrecht-j sorry but this is not my field of expertise. ;-)

jeffreyssmith2nd:
Hey @albrecht-j, I'm the right person for this. My apologies, I just lost track of this PR. I will get it reviewed before the next release.

albrecht-j:
Hey @jeffreyssmith2nd 
any updates to this PR? If we need to change something please let us know :)

albrecht-j:
Hi @jeffreyssmith2nd,

do you have a schedule for the next release?

tisonkun:
Seems a function duplicatie to https://github.com/influxdata/influx-cli/pull/530.

But the InfluxDB team doesn't care. I can have my fork but perhaps switch to some SDK.

I remember Python SDK v1 has such path. Not sure if v2 has.

api/configuration_default.go
```diff
@@ -27,6 +27,15 @@ func NewAPIConfig(params ConfigParams) *Configuration {
 	apiConfig.Scheme = params.Host.Scheme
 	apiConfig.UserAgent = params.UserAgent
 	apiConfig.HTTPClient = &http.Client{Transport: clientTransport}
+
+	if params.Host.Path != "" {
+		// initialize api server configurations with path of host  url
+		apiConfig.Servers = ServerConfigurations{{URL: params.Host.Path}}
+		for key := range apiConfig.OperationServers {
+			apiConfig.OperationServers[key] = ServerConfigurations{{URL: params.Host.Path}}
```
gwossum, line 35: Overall, this looks relatively straightforward and like it will address the issue. 

My one reservation is that it completely overrides the existing `apiConfig.OperationServers[key]` value. We don't lose any valuable information today (just a non-useful description), but I'm concerned this could introduce subtle issues in the future if more valuable information is already populated in the `ServerConfigurations`.

Would it be possible to set the `URL` value without overwriting everything else?

tisonkun: @gwossum Just to handle the path part, you can take a look at #530 which do a light prepend on constructing the final request.

mapero:
Hi, what is the current state of this PR?

Thanks.

Issue #452, opened 2022-10-23
Feature Request: Support specifying path prefix for InfluxDB URL


yoshigev:
I'm opening a feature request for feature that was already done on 1.8: https://github.com/influxdata/influxdb/issues/10294
It was done by @bthesorceror on https://github.com/influxdata/influxdb/pull/16599.

The idea is to allow specifying a sub-path on the URL, so the client could work when an HTTP reverse proxy is routing the requests based on the path. For example, using a URL: `https://example.com/influxdb/`.

Apparently, this support was removed on version 2, as the flags no longer exists.

Is it possible to support this?

Thanks,
Yehoshua

Fuseteam:
Seems related to https://github.com/influxdata/influxdb/issues/15721

jdeus:
Any news?

tisonkun:
Even we change the `apiConfig.Host = params.Host.Host` to use whatever user passes in.

tisonkun:
But it related to some generated code logic. So I can hardly help. Basically we pass the "host" as "https://example.com/influxdb/" and the path part should be conveyed to the final client config and it should work.

tisonkun:
But I anyway made it. Here is a patch and I verify it locally.

PR #530, opened 2024-03-11
feat: support switching basepath

tisonkun:
This closes https://github.com/influxdata/influx-cli/issues/452.

### DIFF
--- a/api/templates/client.mustache
+++ b/api/templates/client.mustache
@@ -294,6 +294,11 @@ func (c *APIClient) prepareRequest(
 		return nil, err
 	}
 
+	// Prepend request base path, if applicable
+    if c.cfg.Path != "" {
+        url.Path = c.cfg.Path + url.Path
+    }
+
 	// Override request host, if applicable
 	if c.cfg.Host != "" {
 		url.Host = c.cfg.Host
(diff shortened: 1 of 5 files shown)


api/templates/client.mustache
```diff
@@ -294,6 +294,11 @@ func (c *APIClient) prepareRequest(
 		return nil, err
 	}
 
+	// Prepend request base path, if applicable
+    if c.cfg.Path != "" {
```
tisonkun, line 298: I can't `make openai` locally so perhaps we still need some formatting here. But the code functions are proper.

```
./etc/generate-openapi.sh
WARNING: The requested image's platform (linux/amd64) does not match the detected host platform (linux/arm64/v8) and no specific platform was requested
Error opening file "/api/contract/openapi/src/legacy/paths/query.yml" 
ENOENT: no such file or directory, open '/api/contract/openapi/src/legacy/paths/query.yml'
make: *** [openapi] Error 1
```

api/templates/client.mustache
```diff
@@ -294,6 +294,11 @@ func (c *APIClient) prepareRequest(
...
+        url.Path = c.cfg.Path + url.Path
```
tisonkun, line 299: Better to handle `/` properly. I don't know if there is an existing utility for it. But users can always self add/remove the `/` on demand.

Issue #533, opened 2024-03-29
Disable password length checking


davidby-influx:
To allow `influxdb` to handle variable password strength validation, disable length checking in the CLI.

See https://github.com/influxdata/influxdb/issues/24856 for details.

PR #534, opened 2024-03-29
fix: disable passord length checks so influxdb validates passwords

davidby-influx:
Disable password length checks and let `influxd` 
do all password strength checks.  

closes https://github.com/influxdata/influx-cli/issues/533


DStrand1 (approved):
LGTM!

PR #536, opened 2024-04-03
fix: handle line protocol wrap with --skipRowOnError

davidby-influx:
closes https://github.com/influxdata/influx-cli/issues/502


DStrand1 (approved):
LGTM!

PR #557, opened 2025-01-30
feat: add `--operator-token` flag to restore subcommand

gwossum:
Add `--operator-token` flag to restore subcommand. If the backup being restored lacks a plaintext operator token, then the token given with `--operator-token` will be used to restore the backup.


davidby-influx (approved):
LGTM

devanbenz (approved):
lgtm

